PR #9, opened 2018-10-08
Add version numbers to dependencies

magnonellie:


omni-viral:
Bors r+

magnonellie:
That was a quick build compared to what happens for the Amethyst repo :open_mouth:

omni-viral:
Yeah. Amethyst takes some time to build.
Aren't tgere some caches to rebuild only crates with changes and their dependencies?

magnonellie:
There are but they don't really work for Amethyst since they get so big. I think the plan is gonna be to move to Microsoft's Azure pipelines.

PR #10, opened 2018-10-16
Add base rendy crate with basic factory implementation

LucioFranco:
This adds the overall `rendy` crate to the project and also adds a
basic `Device` and `Factory` struct.

Opening this for initial review and thoughts! Still, some more work to do.


resource/src/resources.rs
```diff
@@ -142,3 +148,12 @@ impl<M: 'static, B: 'static, I: 'static> Resources<M, B, I> {
         }
     }
 }
+
+impl<M: 'static, B: 'static, I: 'static> Default for Resources<M, B, I> {
```
omni-viral: `'static` bounds seem unnecessary here.
Would you kindly use `derivative` crate instead of manual implementation where possible, thank you.

src/factory.rs
```diff
@@ -0,0 +1,46 @@
...
+impl<D> Factory<D>
+where
+    D: Device,
+{
+    pub fn new<P, H>(device: D, types: P, heaps: H) -> Self
+    where
+        P: IntoIterator<Item = (Properties, u32, Config)>,
+        H: IntoIterator<Item = u64>,
+    {
+        let heaps = unsafe { Heaps::new(types, heaps) };
```
omni-viral: Are you sure about this unsafe block?

src/factory.rs
```diff
@@ -0,0 +1,46 @@
...
+
+        Factory {
+            device,
+            families,
+            heaps,
+            resources: Resources::new(),
+        }
+    }
+
+    pub fn families(&self) -> (&D, &Families<D::CommandQueue>) {
```
omni-viral: You don't need to return `&D` from every method.

src/factory.rs
```diff
@@ -0,0 +1,46 @@
...
+        (&self.device, &self.families)
+    }
+
+    pub fn resources(&self) -> (&D, &Resources<D::Memory, D::Buffer, D::Image>) {
```
omni-viral: `Factory` must encapsulate `Resources` and `Heaps` providing a way to create resources, allocate memory and create memory bound resources.

src/factory.rs
omni-viral: Also you'd need mutable references to families to use them.

Cargo.toml
```diff
@@ -7,5 +7,15 @@ members = [
     "command",
     "layout",
     "layout-derive",
-    "graph",
+    "graph"
 ]
+
+[package]
```
omni-viral: Please, put this crate into its own directory.

Cargo.toml
```diff
@@ -7,5 +7,15 @@ members = [
     "command",
     "layout",
     "layout-derive",
-    "graph",
+    "graph"
```
omni-viral: Should you really remove this comma?

rendy/src/factory.rs
```diff
@@ -0,0 +1,95 @@
...
+{
+    pub fn new<P, H>(device: D, types: P, heaps: H) -> Self
+    where
+        P: IntoIterator<Item = (Properties, u32, Config)>,
+        H: IntoIterator<Item = u64>,
+    {
+        // TODO: make sure this is safe
+        let heaps = unsafe { Heaps::new(types, heaps) };
+
+        let families = Families::new();
```
LucioFranco: I'm also assuming this needs to be created in a different way, or passed in?

rendy/src/lib.rs
```diff
@@ -0,0 +1,15 @@
+extern crate rendy_command;
+extern crate rendy_memory;
+extern crate rendy_resource;
+
+#[cfg(feature = "hal")]
+extern crate gfx_hal as hal;
+
+#[cfg(feature = "ash")]
+extern crate ash;
```
Rhuagh: pub extern?

rendy/src/lib.rs
```diff
@@ -0,0 +1,15 @@
+extern crate rendy_command;
+extern crate rendy_memory;
+extern crate rendy_resource;
+
+#[cfg(feature = "hal")]
+extern crate gfx_hal as hal;
```
Rhuagh: pub extern?

rendy/src/factory.rs
```diff
@@ -0,0 +1,95 @@
...
+where
+    D: Device,
+{
+    pub fn new<P, H>(device: D, types: P, heaps: H) -> Self
+    where
+        P: IntoIterator<Item = (Properties, u32, Config)>,
+        H: IntoIterator<Item = u64>,
+    {
+        // TODO: make sure this is safe
+        let heaps = unsafe { Heaps::new(types, heaps) };
```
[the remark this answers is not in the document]
omni-viral: Unfortunately, you can't
This method is inherently unsafe

rendy/src/factory.rs
omni-viral: Device and queues are created together.
Safe initialization method would create them and query memory types and heaps.

rendy/src/factory.rs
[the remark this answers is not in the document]
LucioFranco: So does that mean, that we need some initialization function that creates types and heaps that are safe to use and therefore making this call safe? Meaning, that this function should be marked as unsafe and should be consumed by a higher level api within Rendy that enforces the safety? Similar to `gfx_render::init`?

rendy/src/factory.rs
LucioFranco: Okay this follows from above then, we need to create some initialization function to handle the safe creation of the Factory. Therefore, I will add families as a parameter to the new function, to allow the initialization function to pass them in.

rendy/src/factory.rs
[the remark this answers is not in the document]
omni-viral: Yeap. Or make this function similar to `gfx_render::init`

rendy/examples/basic.rs
```diff
@@ -0,0 +1,46 @@
...
+
+    //let _factory = rendy::init::<HalDevice, PickFirst, backend::Backend>(config);
+
+    Ok(())
+}
+
+struct PickFirst;
+impl QueuesPicker for PickFirst {
+    fn pick_queues(&self) -> Result<(FamilyId, u32), ()> {
+        Ok((FamilyId(0), 1))
```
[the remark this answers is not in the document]
LucioFranco: Yup this, was just here to mock out this operation, I am planning on adding some to the main crate.

PR #14, opened 2018-11-04
What if Hal

omni-viral:
This change is [Reviewable](https://reviewable.io/reviews/rustgd/rendy/14)


torkleyy:
Can you explain what this PR is about?

omni-viral:
This PR is for comparison. Yet it is not convenient way to compare. So I close. Look for PR in my fork.

PR #18, opened 2018-11-30
Implement create_surface for dx12 in wsi

kabergstrom:


omni-viral:
This PR contains fatal error. `dx12` goes second after `empty` :)

PR #20, opened 2018-12-05
create swapchain_config with default window size

ibloat:
On Wayland (using the gfx vulkan backend) `capabilities.current_extent` can be `None`, which causes [this unwrap](https://github.com/rustgd/rendy/blob/8eecae13a3c75e927e412b074b27e2521e8316a5/wsi/src/lib.rs#L171) to fail.

Looking at https://github.com/gfx-rs/gfx/issues/2421#issuecomment-423852078 and https://github.com/gfx-rs/gfx/pull/2484 the solution seems to be to default to the window dimensions (which `SwapchainConfig::from_caps` does)

### DIFF
--- a/wsi/src/lib.rs
+++ b/wsi/src/lib.rs
@@ -163,28 +163,33 @@ where
         log::info!("Surface capabilities: {:#?}. Pick {} images", capabilities.image_count, image_count);
         assert!(capabilities.usage.contains(usage), "Surface supports {:?}, but {:?} was requested");
 
-        let hidpi_factor = self.window.get_hidpi_factor();
-        let (window_width, window_height) = self.window
-            .get_inner_size()
-            .unwrap()
-            .to_physical(hidpi_factor)
-            .into();
+        let extent = capabilities.current_extent.unwrap_or({
+            let hidpi_factor = self.window.get_hidpi_factor();
+            let start = capabilities.extents.start;
+            let end = capabilities.extents.end;
+            let (window_width, window_height) = self.window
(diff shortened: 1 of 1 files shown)


termhn:
I think it would be better to instead of just *always* using the window dimensions, only use them when `capabilities.current_extent` is `None` i.e.

```rs
       let (swapchain, backbuffer) = device.create_swapchain(
            &mut self.raw,
            gfx_hal::SwapchainConfig {
                present_mode,
                format,
                extent: capabilities.current_extent.unwrap_or({
                    let hidpi_factor = self.window.get_hidpi_factor();
                    let (window_width, window_height) = self.window
                        .get_inner_size()
                        .unwrap()
                        .to_physical(hidpi_factor)
                        .into();
                        gfx_hal::window::Extent2D {
                            width: window_width,
                            height: window_height,
                        }
                }),
                image_count,
                image_layers: 1,
                image_usage: usage,
            },
            None,
        )?;
```

termhn:
Oh! Didn't realize that... Well in that case I honestly don't think it's worth it to duplicate functionaly--I'm in favor of just going back to the original way you had it actually 😅

Thoughts @omni-viral ?

omni-viral:
Seems. Nice.
Take note that this function can fail if requested extent wouldn't match actual surface size.
Which might change right after we queried it.

termhn:
ah, seems bors broke by 2018 edition... making 2018 migration pr now and you can rebase on that hopefully

PR #25, opened 2018-12-19
Cleanup

omni-viral:
This change is [Reviewable](https://reviewable.io/reviews/rustgd/rendy/25)


termhn:
Sorry for delay, I'll try to review this more in depth later tonight

rendy/examples/quads/main.rs
```diff
@@ -171,7 +171,7 @@ where
             let mut rng = rand::thread_rng();
             let uniform = rand::distributions::Uniform::new(0.0, 1.0);
 
-            #[repr(C)] struct PosVel { pos: [f32; 2], vel: [f32; 2], }
+            #[repr(C)] #[derive(Copy, Clone)] struct PosVel { pos: [f32; 2], vel: [f32; 2], }
```
termhn: Better to move these attrs to their own lines before the struct declaration maybe? Not sure what rustfmt says about it

graph/src/node/present.rs
```diff
@@ -87,13 +87,14 @@ where
...
         let target = factory.create_target(self.surface, 3, gfx_hal::image::Usage::TRANSFER_DST)?;
-        let mut pool = factory.create_command_pool(family, ())?;
+        let mut pool = factory.create_command_pool(family)?;
 
         let per_image = match target.backbuffer() {
             gfx_hal::Backbuffer::Images(target_images) => {
                 let buffers = pool.allocate_buffers(PrimaryLevel, target_images.len());
                 target_images.iter().zip(buffers).map(|(target_image, initial)| {
-                    let mut encoder = initial.begin(MultiShot(SimultaneousUse), ());
+                    let mut recording = initial.begin::<MultiShot<_>, _>();
```
termhn: This occurs in several other places too, perhaps change this to `recorder` or even just `cmd_buf` as `recording` doesn't really make sense on its own as an object

omni-viral: I name buffer bindings after their state. This buffer is in **recording** state. It was created from buffer in **initial** state.

factory/src/factory.rs
```diff
@@ -1,12 +1,37 @@
...
-    command::{families_from_device, Family, Reset, CommandPool},
+    command::{families_from_device, Family, Reset, CommandBuffer, CommandPool, Transfer, IndividualReset, PendingState, InitialState, RecordingState, OneShot, PrimaryLevel},
     memory::{Block, Heaps, Write},
     resource::{buffer::{self, Buffer}, image::{self, Image}, Resources},
     wsi::{Surface, Target},
     config::{Config, HeapsConfigure, QueuesConfigure},
 };
 
+#[derive(Debug)]
+struct PendingUploads<B: gfx_hal::Backend> {
```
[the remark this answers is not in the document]
termhn: Right, that's why I could go either way... maybe better to just leave as is, just felt a bit weird as a name.

graph/src/node/present.rs
termhn: Right, I just don't like the name of `recording` as an object. I'd suggest `recording_buf` but it seems a bit long to be typing out everywhere.

factory/src/factory.rs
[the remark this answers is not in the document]
omni-viral: This is still unfinished. I just caught myself thinking the code became too dirty in few places.

graph/src/node/present.rs
termhn: @omni-viral perhaps `cbuf_initial` and `cbuf_recording` then?

command/src/buffer/submit.rs
```diff
@@ -47,32 +37,40 @@ pub unsafe trait Submittable<B: gfx_hal::Backend, L = PrimaryLevel, P = OutsideR
     fn family(&self) -> gfx_hal::queue::QueueFamilyId;
 
     /// Get raw command buffer.
-    fn raw(&self) -> &B::CommandBuffer;
+    /// The command buffer is returned as raw pointer
+    /// because its lifetime is not tied to `Submittable` instance lifetime
+    /// but rather to original `CommandBuffer`.
+    /// The command buffer cannot be freed before commands are complete
```
termhn: Put the notes about safety (this line and after) in #Safety block

omni-viral:
bors r=termhn

PR #26, opened 2018-12-28
Use gfx_hal 0.1.0

64:
`gfx_hal` has finally been published, so I updated rendy to use it. It requires adding unsafe blocks in a number of places.

### DIFF
--- a/chain/Cargo.toml
+++ b/chain/Cargo.toml
@@ -5,6 +5,6 @@ authors = ["omni-viral <scareaangel@gmail.com>"]
 edition = "2018"
 
 [dependencies]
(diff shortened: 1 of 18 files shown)


termhn:
Thanks! LGTM. Is this `rustfmt`'d?

64:
Sorry, it's not. I seemed to remember the other PR doing a rustfmt run, but I guess I just imagined that. I'll do it now.

64:
Oh yeah, the diff is huge - it affects way more files than I've touched. Want me to commit it anyway?

termhn:
Uhh, might as well just do a rustfmt commit later then and leave this as is.

PR #29, opened 2018-12-31
Fix surface creation on iOS

Moxinilian:
This reimplements surface creation on iOS thanks to the unified `create_surface`.


omni-viral:
Thanks!
bors r+

PR #33, opened 2019-01-07
Simpler introduction text

erlend-sh:
An attempt to clarify what Rendy does.

To go along with this change I suggest editing the GitHub repo description of "Yet another Vulkan based rendering engine." to simply say: A rendering engine based on`gfx-hal`


omni-viral:
@erlend-sh Seems fine! Thanks!

bors r+

PR #37, opened 2019-01-11
Views and samplers

StarArawn:
**Don't merge this yet.** 
I think there are some things we can do better.
I'm not convinced the hash map is the best way to cache the samplers. Right now I'm mainly just looking for feedback, and hopefully when this is cleaned up and looking good we can get it in.

I'll also be adding a quad/sprite example to showcase how to use textures. A lot of this will probably have to be updated once DescriptorSet stuff is figured out.

Edit: Based on feedback I've changed a lot of how this works so the above stuff isn't all relevant anymore.  
Things to still figure out:
`ImageView`'s are attached to `Image`'s. We need to make sure that the `Image` isn't dropped if `ImageView` is still using it. Also we have a lot of issues when the app attempts to dispose of everything. Not sure why.


factory/src/factory.rs
```diff
@@ -1,8 +1,9 @@
 
+use rendy_resource;
```
omni-viral: This is unused.

rendy/examples/triangle/main.rs
```diff
@@ -15,7 +15,6 @@
 #![deny(unconditional_recursion)]
 #![deny(unions_with_drop_fields)]
 #![deny(while_true)]
-#![deny(unused)]
```
omni-viral: Would you kindly put that back ;)
If anything change to `warn` to not forget it later.

resource/src/image/mod.rs
```diff
@@ -145,3 +145,76 @@ where
...
+#[derive(Debug)]
+pub struct ImageView<B: gfx_hal::Backend> {
+    escape: Escape<InnerView<B>>,
+    info: ViewInfo,
+}
+
+#[doc(hidden)]
+#[derive(Debug)]
+pub struct InnerView<B: gfx_hal::Backend> {
+    raw: B::ImageView,
```
omni-viral: Here we need to put `KeepAlive` taken from `Image`.

resource/src/resources.rs
```diff
@@ -176,6 +180,67 @@ where
...
+    }
+
+    /// Create a sampler.
+    pub fn create_sampler(
+        &mut self,
+        device: &impl gfx_hal::Device<B>,
+        filter: gfx_hal::image::Filter,
+        wrap_mode: gfx_hal::image::WrapMode,
+    ) -> Result<&Sampler<B>, failure::Error> {
+        Ok(self.sampler_cache.get(device, filter, wrap_mode).unwrap())
```
omni-viral: Sampler should be returned by value. Returning by references lock whole `Resources` and transitively whole `Factory`. Probably `Sampler` should be clonable.

resource/src/sampler/cache.rs
```diff
@@ -0,0 +1,82 @@
+//! A cache to store and retrieve samplers
+use std::collections::HashMap;
+use std::collections::hash_map::Entry;
+use super::{Sampler, Info};
+
+#[doc(hidden)]
+#[derive(Debug)]
+pub struct SamplerCache<B: gfx_hal::Backend> {
+    // TODO: figure out way to store this better. Perhaps we can clone/copy sampler so we don't need a reference?
+    samplers: HashMap<gfx_hal::image::Filter, HashMap<gfx_hal::image::WrapMode, Sampler<B>>>,
```
omni-viral: Better merge the key and make it `HashMap<(Filter, WrapMode), Sampler<B>>`

resource/src/sampler/cache.rs
```diff
@@ -0,0 +1,82 @@
...
+    B: gfx_hal::Backend,
+{
+    #[doc(hidden)]
+    pub fn get(
+        &mut self,
+        device: &impl gfx_hal::Device<B>,
+        filter: gfx_hal::image::Filter,
+        wrap_mode: gfx_hal::image::WrapMode
+    ) -> Option<&Sampler<B>> {
+        match self.samplers.entry(filter) {
```
omni-viral: This pattern is very common so there is a function for it
`self.samplers.entry((filter, wrap_mode)).or_insert_with(|| Self::create(device, filter, wrap_mode)`

resource/src/resources.rs
omni-viral: Better have unwraps close to code that ensures they'll never panic.

texture/src/texture.rs
```diff
@@ -2,22 +2,26 @@
...
+    resource::sampler::Sampler,
     factory::{Factory, ImageState},
     util::cast_cow,
 };
 
 /// Static image.
 /// Can be loaded from various of formats.
 #[derive(Debug)]
-pub struct Texture<B: gfx_hal::Backend> {
+pub struct Texture<'a, B: gfx_hal::Backend> {
```
omni-viral: See how lifetimes are viral. They infect all types. Better not use them freely.

resource/src/image/mod.rs
jmitchell89: Wait don't I already use keep alive here:
```
    pub fn keep_alive(&self) -> KeepAlive {
        Escape::keep_alive(&self.escape)
    }
```

StarArawn: Wait don't I already use keep alive here:
```
    pub fn keep_alive(&self) -> KeepAlive {
        Escape::keep_alive(&self.escape)
    }
```

resource/src/resources.rs
StarArawn: If `resource:Sampler` is clone-able that means `B::Sampler` should also be clone-able but it is not.

StarArawn: Fixed this issue by using Escape.

texture/src/texture.rs
```diff
@@ -85,13 +90,25 @@ impl<'a> TextureBuilder<'a> {
         self
     }
 
+    /// With image view kind.
+    pub fn with_view_kind(mut self, view_kind: gfx_hal::image::ViewKind) -> Self {
```
StarArawn: TODO: `Kind` and `ViewKind` are very similar. Perhaps we can just figure `Kind` from what `ViewKind` it is?

texture/src/texture.rs
StarArawn: Thankfully it is gone now! :tada:

resource/src/sampler/cache.rs
StarArawn: Done!

rendy/examples/triangle/main.rs
StarArawn: Oops it should be back now!

resource/src/image/mod.rs
StarArawn: I think I need you to elaborate a little more about how KeepAlive works. I took a look at the comment doc but it's not very clear to me where `image.keep_alive()` should be called.

resource/src/sampler/cache.rs
StarArawn: A tuple makes much more sense. 👍

resource/src/image/mod.rs
omni-viral: When you create this `ImageView`

texture/src/texture.rs
omni-viral: Not entirely. View kind must be compatible with image kind. See this table
https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/VkImageViewCreateInfo.html

resource/src/resources.rs
```diff
@@ -176,6 +180,67 @@ where
...
+            range: &'a dyn std::fmt::Debug,
+        };
+        log::trace!("{:#?}", CreateImageView {
+            image: &image,
+            view_kind: &view_kind,
+            format: &format,
+            swizzle: &swizzle,
+            range: &range,
+        });
+
```
omni-viral: To make this function safe we should follow valid usage guidelines
https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/vkCreateImageView.html

StarArawn: I guess that means making sure the info is valid:
https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/VkImageViewCreateInfo.html

resource/src/escape.rs
```diff
@@ -78,6 +78,14 @@ pub struct Escape<T> {
     inner: std::sync::Arc<Inner<T>>,
 }
 
+impl<T> Clone for Escape<T> {
```
omni-viral: `Escape` can't implement `Clone` because it allows mutable access to content hence safe code can create two `&mut T` pointing to the same value.

I see two possible solutions here
* `clone` can be made as unsafe function of the type
* Another version of `Escape` (`EscapeShared` IDK, I'm bad at naming) made that doesn't permit mutable access to escaping value and implement `Clone`

I tend to the second option.

jmitchell89: I think the `EscapeShared` makes sense. How do we make the arc inside of escape immutable?

StarArawn: I think the `EscapeShared` makes sense. How do we make the arc inside of escape immutable?

omni-viral: It is so by default. In `Escape` I use `UnsafeCell` to get mutable pointer for shared reference.
In `EscapeShader` it is basically not required :)

texture/src/texture.rs
```diff
@@ -102,6 +102,29 @@ impl<'a> TextureBuilder<'a> {
...
+            },
+            gfx_hal::image::Kind::D2(..) => {
+                match self.view_kind {
+                    gfx_hal::image::ViewKind::D2 | gfx_hal::image::ViewKind::D2Array => true,
+                    _ => false,
+                }
+            },
+            gfx_hal::image::Kind::D3(..) => {
+                match self.view_kind {
+                    gfx_hal::image::ViewKind::D2 | gfx_hal::image::ViewKind::D2Array | gfx_hal::image::ViewKind::D3 => true,
```
omni-viral: Here `ViewKind::D2` and `ViewKind::D2Array` only match if image was created with appropriate flag.

texture/src/texture.rs
```diff
@@ -113,6 +136,8 @@ impl<'a> TextureBuilder<'a> {
     where
         B: gfx_hal::Backend,
     {
+        assert!(self.match_kind());
```
omni-viral: Better do this check in `Resources::create_image_view`

texture/src/texture.rs
```diff
@@ -85,13 +92,37 @@ impl<'a> TextureBuilder<'a> {
...
+    }
+
     /// Build texture.
     pub fn build<B>(
         &self,
         queue: QueueId,
         access: gfx_hal::image::Access,
         layout: gfx_hal::image::Layout,
-        factory: &mut Factory<B>,
+        factory: &'a mut Factory<B>,
```
omni-viral: I believe you don't need this lifetime.
In fact, if `'a` is `'static` you won't be able to call this function at all unless `'a` could be shrinked.

omni-viral:
@StarArawn could you rebase please.

omni-viral:
Changes were merged in bigger PR #45

PR #41, opened 2019-01-18
Added a function to get the picked format from a surface.

StarArawn:
Added a `get_surface_format` to get the picked surface format to use for `graph_builder.create_image`.


omni-viral:
Could you please fix building errors? )

PR #43, opened 2019-02-03
Refactor render node system

omni-viral:


graph/src/node/render/pass.rs
```diff
@@ -0,0 +1,666 @@
...
+        (
+            Layout::DepthStencilAttachmentOptimal,
+            Layout::DepthStencilReadOnlyOptimal,
+        ) => Layout::DepthStencilAttachmentOptimal,
+        (_, _) => Layout::General,
+    }
+}
+
+
+#[cfg(feature = "metal")]
```
termhn: why do we need the cfg separation plus the typeid checking? Couldn't we just use cfg or typeid and not both?

graph/src/node/present.rs
```diff
@@ -31,57 +31,77 @@ where
...
-    /// Create present builder
-    pub fn new(
-        surface: Surface<B>,
-    ) -> Self {
-        PresentDesc {
-            surface,
-        }
+    /// Add dependency.
+    /// Node will be placed after its dependencies.
+    pub fn add_dependency(&mut self, dependency: NodeId) -> &mut Self {
```
termhn: Why do these need to be directly on the PresentBuilder instead of in NodeBuilder trait?

graph/src/node/render/pass.rs
omni-viral: Code under `#[cfg(feature = "backendname")]` compiles when that backend is enabled, but more than one backend can be enabled simultaneously, we should also check what backend `B` is.
Without backend feature you can't refer to it, so check if `B` is `gfx_backend_metal::Backend` won't compile as there is no `gfx_backend_metal` crate.

graph/src/node/present.rs
omni-viral: You actually bring me to another mistake I made. `SubpassBuilder` should support adding dependencies. However it doesn't make sense to add dependencies to render pass (`RenderPassNodeBuilder` implements `NodeBuilder`) as it is just collection of subpasses and its dependencies is just dependencies of all subpasses.

omni-viral:
> assuming that stuff is correct

That's big assumption 😁

termhn:
Heh, well, I didn’t see anything *glaringly obvious* but with this sort of work that doesn’t mean much ;p

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

graph/src/node/render/pass.rs
termhn: Fair enough

PR #50, opened 2019-02-14
consume RenderGroup builder on build

Frizi:


graph/src/node/render/group/simple.rs
```diff
@@ -190,30 +196,43 @@ pub struct SimpleRenderGroup<B: Backend, P> {
     pipeline: P,
 }
 
-impl<B, T, P> RenderGroupDesc<B, T> for P
+#[derive(Derivative)]
+#[derivative(Debug)]
+pub struct SimpleRenderGroupDesc<B, T, P>
```
omni-viral: Can you remove `B` and `T` parameters?

Frizi: I need them for `SimpleGraphicsPipelineDesc` bound

PR #52, opened 2019-02-17
Add Swapchain image count and present mode configuration to PresentBuilder

termhn:


rendy/examples/meshes/main.rs
```diff
@@ -435,12 +417,55 @@ fn main() {
             .into_pass(),
     );
 
-    graph_builder.add_node(PresentNode::builder(surface, color).with_dependency(pass));
+    let present_builder = PresentNode::builder(surface, factory.physical(), color)
+        .with_dependency(pass);
+
+    let frames = present_builder.image_count() as usize;
```
omni-viral: That's not the the number of frames Graph renders in parallel.

wsi/src/lib.rs
```diff
@@ -159,26 +159,38 @@ where
             .unwrap()
     }
 
+    /// Get surface compatibility
+    pub fn compatibility(
```
omni-viral: Mark this function unsafe. `PhysicalDevice` must belong to the same `Instance` as the `Surface`
Add safe `compatibility` function to `Factory` that will call this one.
Later we will add assertion there that surface was created from `Factory`'s `Instance`.

graph/src/graph/mod.rs
```diff
@@ -363,7 +369,7 @@ where
                 .filter_map(|x| x)
                 .map(|(image, _)| image)
                 .collect(),
-            inflight: 3,
+            inflight: self.frames_in_flight as _,
```
omni-viral: Possible loss of data.
Use the same integer type for `Graph` and `GraphBuilder`.
I guess `usize` is nice.

graph/src/node/present.rs
```diff
@@ -75,6 +110,60 @@ where
...
+    /// indicates a higher prioirity for that mode.
+    /// 
+    /// ## Panics
+    /// - Panics if none of the provided `PresentMode`s are supported.
+    pub fn with_present_modes_priority<PF>(mut self, present_modes_priority: PF) -> Self 
+        where PF: Fn(gfx_hal::PresentMode) -> Option<usize>
+    {
+        if !self.present_modes_caps.iter().any(|m| {
+            match present_modes_priority(*m)  {
+                Some(_) => true,
```
omni-viral: `present_modes_priority(*m).is_some()`

omni-viral:
Thanks! There just few nits.

graph/src/graph/mod.rs
termhn: I mean there's no way you're gonna want more than a u32's worth of frames_in_flight, so... x)

omni-viral: Probably, until year 3019 :)
We can even settle on `u8`

factory/src/factory.rs
```diff
@@ -252,32 +258,72 @@ where
...
+    pub fn get_surface_compatibility(
+        &self,
+        surface: &Surface<B>
+    ) -> (
+        gfx_hal::window::SurfaceCapabilities,
+        Option<Vec<gfx_hal::format::Format>>,
+        Vec<gfx_hal::PresentMode>,
+        Vec<gfx_hal::CompositeAlpha>
+    ) {
+        #[cfg(not(feature = "no-slow-safety-checks"))]
```
omni-viral: I'd add `rendy_slow_assert` macro that is no-op under "no-slow-safety-checks" feature.
Just to be sure no one forgot to add feature on top of the slow check.
Similar to how you use `debug_assert!(test)` instead of `#[cfg(feature = "debug-assertions")] assert!(test)`

factory/src/factory.rs
```diff
@@ -633,16 +679,18 @@ where
         log::info!("Queues: {:#?}", get_queues);
 
         let Gpu { device, mut queues } = unsafe { adapter.physical_device.open(&create_queues) }?;
+        
+        let id = FACTORY_ID.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
 
         assert_eq!(
```
omni-viral: This assertion should be removed I guess.

wsi/src/lib.rs
```diff
@@ -159,26 +176,40 @@ where
...
-            .iter()
-            .max_by_key(|mode| match mode {
-                gfx_hal::PresentMode::Immediate => 0,
-                gfx_hal::PresentMode::Mailbox => 3,
-                gfx_hal::PresentMode::Fifo => 2,
-                gfx_hal::PresentMode::Relaxed => 1,
-            })
-            .unwrap();
+        if !present_modes.contains(&present_mode) {
+            panic!(
```
omni-viral: Shouldn't this return error instead of panicking?

termhn: Probably

factory/src/factory.rs
termhn: Good idea!

PR #53, opened 2019-02-18
Allow storing subpass in rendergroup

Frizi:
I need to (indirectly) store subpass in the rendergroup, because my pipeline defs has to be created lazily based on world state. I coudn't do that, because `RenderGroup` has to be `'static` right now.

EDIT: ~~This is potentially unsound, got a weird warning that should be a compilation error.~~ (should be fixed now)

### DIFF
--- a/graph/src/node/render/group/mod.rs
+++ b/graph/src/node/render/group/mod.rs
@@ -62,10 +62,17 @@ pub trait RenderGroup<B: Backend, T: ?Sized>: std::fmt::Debug + Send + Sync {
         factory: &Factory<B>,
         queue: QueueId,
         index: usize,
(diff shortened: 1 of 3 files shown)


omni-viral:
Since `Subpass` owns `RenderGroup`s wouldn't this mean you want self-reference?

omni-viral:
You would have to pin `RenderPassNode`.

Frizi:
Well, i'm not sure I need that. I woudn't have to store the Subpass ref if I could retrieve it during `prepare`.

omni-viral:
You totally can add another parameter

Frizi:
Parameter added

PR #55, opened 2019-02-23
Use core::mem::swap instead of std::mem::replace on PresentNode acquire semaphores

termhn:
Gets rid of need to use Option and makes no difference in effect

### DIFF
--- a/graph/src/node/present.rs
+++ b/graph/src/node/present.rs
@@ -31,7 +31,7 @@ struct ForImage<B: gfx_hal::Backend> {
 #[derive(Debug)]
 pub struct PresentNode<B: gfx_hal::Backend> {
     per_image: Vec<ForImage<B>>,
-    free: Option<B::Semaphore>,
+    free_acquire: B::Semaphore,
     target: Target<B>,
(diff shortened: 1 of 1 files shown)


graph/src/node/present.rs
```diff
@@ -303,11 +303,10 @@ where
         signals: &[&'a B::Semaphore],
         fence: Option<&mut Fence<B>>,
     ) {
-        let acquire = self.free.take().unwrap();
-        let next = self.target.next_image(&acquire).unwrap();
+        let next = self.target.next_image(&self.free_acquire).unwrap();
         log::trace!("Present: {:#?}", next);
         let ref mut for_image = self.per_image[next[0] as usize];
-        self.free = Some(std::mem::replace(&mut for_image.acquire, acquire));
+        core::mem::swap(&mut for_image.acquire, &mut self.free_acquire);
```
omni-viral: Why not `std::mem::swap`?

termhn: `std::mem::swap` is just a reexport of `core::mem::swap`, no reason in particular tho

PR #59, opened 2019-03-01
Add missing serde imports for Serialize and Deserialize

Frizi:


factory/src/config.rs
```diff
@@ -54,7 +54,8 @@ unsafe impl QueuesConfigure for OneGraphicsQueue {
 
 /// Saved config for queues.
 #[derive(Clone, Debug)]
-#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+// TODO: Enable serde when FamilyId will be serializable
```
omni-viral: Could you create PR to the `gfx-hal` with that?

Frizi: Will do!

PR #63, opened 2019-03-06
Use serde_bytes in mesh builder

Frizi:
This speeds up serialization/deserialization in asset pipeline by some crazy numbers, something like 100x for bigger meshes.

### DIFF
--- a/mesh/src/mesh.rs
+++ b/mesh/src/mesh.rs
@@ -83,9 +83,25 @@ impl<'a> From<Cow<'a, [u32]>> for Indices<'a> {
 #[derive(Clone, Debug)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct MeshBuilder<'a> {
-    vertices: smallvec::SmallVec<[(Cow<'a, [u8]>, VertexFormat<'static>); 16]>,
-    indices: Option<(Cow<'a, [u8]>, gfx_hal::IndexType)>,
+    vertices: smallvec::SmallVec<[RawVertices<'a>; 16]>,
+    indices: Option<RawIndices<'a>>,
     prim: gfx_hal::Primitive,
+}
+
+#[derive(Clone, Debug)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+struct RawVertices<'a> {
+    #[cfg_attr(feature = "serde", serde(with = "serde_bytes"))]
+    vertices: Cow<'a, [u8]>,
+    format: VertexFormat<'static>,
+}
+
+#[derive(Clone, Debug)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+struct RawIndices<'a> {
+    #[cfg_attr(feature = "serde", serde(with = "serde_bytes"))]
+    indices: Cow<'a, [u8]>,
+    index_type: gfx_hal::IndexType,
 }
 
 impl<'a> MeshBuilder<'a> {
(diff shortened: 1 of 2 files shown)


mesh/src/mesh.rs
```diff
@@ -83,11 +83,27 @@ impl<'a> From<Cow<'a, [u32]>> for Indices<'a> {
...
-    vertices: smallvec::SmallVec<[(Cow<'a, [u8]>, VertexFormat<'static>); 16]>,
-    indices: Option<(Cow<'a, [u8]>, gfx_hal::IndexType)>,
+    vertices: smallvec::SmallVec<[VerticesDef<'a>; 16]>,
+    indices: Option<IndicesDef<'a>>,
     prim: gfx_hal::Primitive,
 }
 
+#[derive(Clone, Debug)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+struct VerticesDef<'a> {
```
omni-viral: Why not just `Vertices`?

mesh/src/mesh.rs
```diff
@@ -83,11 +83,27 @@ impl<'a> From<Cow<'a, [u32]>> for Indices<'a> {
...
+    #[cfg_attr(feature = "serde", serde(with = "serde_bytes"))]
```
omni-viral: [serde_bytes::deserialize](https://github.com/serde-rs/bytes/blob/master/src/lib.rs#L111) (which is used by generated code thanks to this attribute) is not optimal for `Cow<'a, [u8]>` type since it forcefully convert data into `Vec<u8>` and produces `Cow::Owned` variant, even if deserializer gives [borrowed byte slice](https://docs.rs/serde/1.0.89/serde/de/trait.Visitor.html#method.visit_borrowed_bytes) or [borrowed string](https://docs.rs/serde/1.0.89/serde/de/trait.Visitor.html#method.visit_borrowed_str).

Frizi: I'm not entirely sure, but we probably want that, as it keeps the bincode serialized buffer unused and it can be discarded after loading. When we would have borrows from it, the whole buffer would have to be kept in memory.

mesh/src/mesh.rs
Frizi: That would probably be fine, i'll change that

mesh/src/mesh.rs
omni-viral: Yes, but whole bincode buffer is just few bytes bigger than vertices buffer.
And `MeshBuilder` isn't going to live for long. It will be discarded once data is in GPU buffers.

mesh/src/mesh.rs
Frizi: Well, i can't use "Vertices" and "Indices", because such names are already used for enum. Will go with "RawVertices" and "RawIndices".

mesh/src/mesh.rs
omni-viral: We probably want to have `to_owned` method that converts all `Cow`s into owned variant with `'static` lifetime. This way we can be most flexible - borrow buffers from deserializer and detach from deserializer lifetime if we need to.

Frizi: That's reasonable, but it would require writing custom deserializer and changing the API. I would prefer to do it in a separate PR

omni-viral: IIRC Despite that nothing is actually borrowed from deserializer, generated `Deserialize` impl sticks `MeshBuilder` lifetime to `Deserializer`'s, just because `MeshBuilder` has a lifetime parameter. So we still keep whole bincode buffer in memory :)

Frizi: That's sad :( Anyway, borrowing combined with `fn to_owned() -> MeshBuilder<'static>` should do the job. Just want to do it separately anyway.

omni-viral:
@Frizi could you fix conflict, please?

Frizi:
@omni-viral done

PR #64, opened 2019-03-07
Update wavefront_obj

Frizi:
The loader also avoids string allocation, because parser can now accept borrowed data.

### DIFF
--- a/mesh/src/format/obj.rs
+++ b/mesh/src/format/obj.rs
@@ -9,8 +9,8 @@ use {
 };
 
 /// Load mesh data from obj.
-pub fn load_from_obj(bytes: Vec<u8>, _: ()) -> Result<MeshBuilder<'static>, failure::Error> {
-    let string = String::from_utf8(bytes)?;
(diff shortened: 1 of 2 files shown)


mesh/src/format/obj.rs
```diff
@@ -10,7 +10,7 @@ use {
 
 /// Load mesh data from obj.
 pub fn load_from_obj(bytes: Vec<u8>, _: ()) -> Result<MeshBuilder<'static>, failure::Error> {
-    let string = String::from_utf8(bytes)?;
+    let string = std::str::from_utf8(&bytes)?;
```
omni-viral: There were no string allocation here FYI )
But now we can make the function take borrowed slice as well.

Frizi: That's actually true 🤦‍♂️  Still, the library was previously reallocating the string inside :P

Frizi: I've changed the function to accept borrowed slice. This will allow us to avoid copying in the asset loader later.

PR #66, opened 2019-03-09
Support texture loading using image crate

Frizi:
This adds a possibility to load various image formats directly into a texture builder.

This code is probably wrong for any texture kind except 2D, because I don't really get the way how layers are treated in memory.

### DIFF
--- a/texture/src/format/image.rs
+++ b/texture/src/format/image.rs
@@ -193,22 +193,20 @@ mod serde_image_format {
         HDR,
     }
 
+    #[derive(Serialize, Deserialize)]
+    struct Helper(#[serde(with = "SerdeImageFormat")] image::ImageFormat);
+
     pub fn serialize<S: Serializer>(
         value: &Option<image::ImageFormat>,
         serializer: S,
     ) -> Result<S::Ok, S::Error> {
-        #[derive(Serialize)]
-        struct Helper<'a>(#[serde(with = "SerdeImageFormat")] &'a image::ImageFormat);
-        value.as_ref().map(Helper).serialize(serializer)
+        value.map(Helper).serialize(serializer)
     }
 
     pub fn deserialize<'de, D: Deserializer<'de>>(
         deserializer: D,
     ) -> Result<Option<image::ImageFormat>, D::Error> {
-        #[derive(Deserialize)]
-        struct Helper(#[serde(with = "SerdeImageFormat")] image::ImageFormat);
-        let helper = Option::deserialize(deserializer)?;
-        Ok(helper.map(|Helper(external)| external))
+        Ok(Option::deserialize(deserializer)?.map(|Helper(format)| format))
     }
 }
 
--- a/texture/src/format/image.rs
+++ b/texture/src/format/image.rs
@@ -23,7 +23,6 @@ pub enum LayerLayout {
     Column,
 }
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
 struct DataLayout {
     /// distance between lines in texels
     pub line_stride: u32,
@@ -49,12 +48,12 @@ impl LayerLayout {
     fn data_layout(&self, image_width: u32, image_height: u32, layers: u32) -> DataLayout {
         match self {
             LayerLayout::Row => DataLayout {
-                line_stride: image_width * layers,
-                layer_stride: image_width,
+                line_stride: image_width,
+                layer_stride: image_width / layers,
             },
             LayerLayout::Column => DataLayout {
                 line_stride: image_width,
-                layer_stride: image_width * image_height,
+                layer_stride: image_width * image_height / layers,
             },
         }
     }
--- a/texture/src/format/image.rs
+++ b/texture/src/format/image.rs
@@ -129,18 +129,19 @@ pub struct ImageTextureConfig {
     filter: gfx_hal::image::Filter,
 }
 
-macro_rules! set_data {
-    ($builder:expr, $repr:expr, $img:expr) => {
+macro_rules! dyn_format {
+    ($channel:ty, $size:ty, $repr:expr) => {{
+        use pixel::{AsPixel, Pixel};
         match $repr {
-            Repr::Unorm => $builder.set_data(img_into_vec::<pixel::Unorm, _>($img)),
-            Repr::Inorm => $builder.set_data(img_into_vec::<pixel::Inorm, _>($img)),
-            Repr::Uscaled => $builder.set_data(img_into_vec::<pixel::Uscaled, _>($img)),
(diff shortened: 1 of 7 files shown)


texture/src/format/image.rs
```diff
@@ -0,0 +1,283 @@
...
+
+    let format = config
+        .format
+        .map_or_else(|| image::guess_format(bytes), |f| Ok(f))?;
+    let image = image::load_from_memory_with_format(bytes, format)?;
+
+    let mut builder = TextureBuilder::new();
+
+    let (w, h) = image.dimensions();
+    builder.set_data_width(w);
```
omni-viral: this is line stride - distance between lines in bytes

texture/src/format/image.rs
```diff
@@ -0,0 +1,283 @@
...
+    builder.set_data_height(h);
```
omni-viral: this is layer stride - distance between layers/planes in bytes

Frizi: should it be (w * h) then? Or is it implicitly multiplied in the driver?

omni-viral: usually it is `line_stride * hight`
where `line_stride = width * pixel_size + line_paddind`

texture/src/format/image.rs
```diff
@@ -0,0 +1,262 @@
...
+        ICO,
+        HDR,
+    }
+
+    pub fn serialize<S: Serializer>(
+        value: &Option<image::ImageFormat>,
+        serializer: S,
+    ) -> Result<S::Ok, S::Error> {
+        #[derive(Serialize)]
+        struct Helper<'a>(#[serde(with = "SerdeImageFormat")] &'a image::ImageFormat);
```
omni-viral: `ImageFormat` is `Copy`, so maybe just copy it?

texture/src/format/image.rs
```diff
@@ -0,0 +1,262 @@
...
+#[derive(Derivative)]
+#[derivative(Default)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct ImageTextureConfig {
+    /// Interpret the image as given format.
+    /// When `None`, format is determined automatically based on magic bytes.
+    /// Automatic method doesn't support TGA format.
+    #[cfg_attr(feature = "serde", serde(with = "serde_image_format"))]
+    format: Option<image::ImageFormat>,
+    repr: Repr,
```
omni-viral: Doesn't image format dictate repr?

texture/src/format/image.rs
```diff
@@ -0,0 +1,262 @@
...
+    let image_format = config
+        .format
+        .map_or_else(|| image::guess_format(bytes), |f| Ok(f))?;
+    let image = image::load_from_memory_with_format(bytes, image_format)?;
+
+    let (w, h) = image.dimensions();
+    let (kind, layout) = config.kind.layout_and_kind(w, h);
+
+    let (vec, format) = match image {
+        DynamicImage::ImageLuma8(img) => (img.into_vec(), dyn_format!(R, _8, config.repr)),
```
omni-viral: We should configure swizzles for this kind of formats.
`Luma` requires swizzle `rrr` and `LumaA` - `rrrg`.

texture/src/format/image.rs
Frizi: Not necessarily. You can still load PNG with non-color data, like normal map. You don't want to load it as Srgb, right?

texture/src/format/image.rs
Frizi: Possibly, but loading it as single channel should work as well IMHO.

Frizi:
Derives added

texture/src/format/image.rs
omni-viral: You need to both, load it as single channel **and** specify swizzle for image view.
Otherwise you'll see terminator vision.

Frizi: Most probably fixed :)

texture/src/format/image.rs
```diff
@@ -0,0 +1,288 @@
...
+        }
+    }
+
+    fn data_layout(&self, image_width: u32, image_height: u32, layers: u32) -> DataLayout {
+        match self {
+            LayerLayout::Row => DataLayout {
+                line_stride: image_width * layers,
+                layer_stride: image_width,
+            },
+            LayerLayout::Column => DataLayout {
```
omni-viral: `image_width` and `image_height` are width and hight of the original image, not of the layer.
Should be
```
DataLayout {
    line_stride: image_width,
    layer_stride: image_height / layers,
}
```

texture/src/format/image.rs
omni-viral: Right.

texture/src/texture.rs
```diff
@@ -57,6 +59,27 @@ impl<'a> TextureBuilder<'a> {
...
+        self
+    }
+
+    /// Set pixel data with manual format definition.
+    pub fn set_raw_data(
+        &mut self,
+        data: impl Into<std::borrow::Cow<'a, [u8]>>,
+        format: gfx_hal::format::Format,
+    ) -> &mut Self {
+        self.data = cast_cow(data.into());
```
omni-viral: `cast_cow` is unnecessary here.

texture/src/format/image.rs
```diff
@@ -0,0 +1,288 @@
...
+
+/// Determines the way layers are being stored in source image.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub enum LayerLayout {
+    Row,
+    Column,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
omni-viral: I meant public types, but ok )

Frizi: Haha 🤦‍♀️

texture/src/format/image.rs
```diff
@@ -0,0 +1,288 @@
...
+    fn layer_height(&self, image_height: u32, layers: u32) -> u32 {
+        match self {
+            LayerLayout::Row => image_height,
+            LayerLayout::Column => image_height / layers,
+        }
+    }
+
+    fn data_layout(&self, image_width: u32, image_height: u32, layers: u32) -> DataLayout {
+        match self {
+            LayerLayout::Row => DataLayout {
```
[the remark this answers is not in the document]
Frizi: 👍

texture/src/format/image.rs
Frizi: 👍

texture/src/texture.rs
Frizi: 👍

PR #68, opened 2019-03-11
Add set_scissors to RenderPassEncoder

bgourlie:
This seemed like a pretty straightforward PR, but I'm relatively new to graphics stuff so apologies if this is missing anything.

### DIFF
--- a/command/src/buffer/encoder.rs
+++ b/command/src/buffer/encoder.rs
@@ -1,4 +1,5 @@
 use {
+    std::borrow::Borrow,
     super::{
         level::{Level, PrimaryLevel, SecondaryLevel},
         state::RecordingState,
@@ -325,6 +326,17 @@ where
                 draw_count,
                 stride,
             )
+        }
+    }
+
+    /// Set scissors
+    pub fn set_scissors<T>(&mut self, first_scissor: u32, rects: T)
+    where
+        T: IntoIterator,
+        T::Item: Borrow<gfx_hal::pso::Rect>,
+    {
+        unsafe {
+            gfx_hal::command::RawCommandBuffer::set_scissors(self.inner.raw, first_scissor, rects)
         }
     }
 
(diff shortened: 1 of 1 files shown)


command/src/buffer/encoder.rs
```diff
@@ -328,6 +329,17 @@ where
         }
     }
 
+    /// Set scissors
+    pub fn set_scissors<T>(&mut self, first_scissor: u32, rects: T)
```
omni-viral: This file uses `impl Trait` in argument position. Please, avoid style inconsistency.

command/src/buffer/encoder.rs
```diff
@@ -328,6 +329,17 @@ where
...
+    where
+        T: IntoIterator,
+        T::Item: Borrow<gfx_hal::pso::Rect>,
```
omni-viral: `Item = &Rect` would do just fine.
Or `Item = &impl Borrow<Rect>` for maximum flexibility.

There is no point of taking ownership of the item if you can only call `Borrow::borrow` that takes a reference.

command/src/buffer/encoder.rs
omni-viral: According to Vulkan spec this function can be called both inside and outside render pass.
Hence this function belongs to `EncoderCommon` with `C: Supports<Graphics>` bound.

bgourlie:
All comments should be addressed :)

command/src/buffer/encoder.rs
```diff
@@ -221,6 +221,17 @@ where
         }
     }
 
+    /// Set scissors
+    pub fn set_scissors<'b>(
+        &mut self,
+        first_scissor: u32,
+        rects: impl IntoIterator<Item = &'b gfx_hal::pso::Rect>,
+    ) {
```
omni-viral: Here we need `where C: Supports<Graphics>` and `self.capability.assert();` on the first line in function body.
Because setting scissors on non-graphics queue is invalid.

bgourlie:
Whoops. Should be good now, assuming the build passes.

PR #72, opened 2019-03-15
[WIP] [FEEDBACK] Shader Reflection implementation

jaynus:
This pull request includes:
 - [x] Working Spirv Reflection for basic example (see quads example)
 - [x] spirv-reflect -> gfx_hall middleware layer for converting the appropriate descriptor and shader information
- [x] rendy_graph includes some feature-gated changes, due to the need to finalize the gfx_hal middlware, see ShaderReflectBuilder
- [x] Changes in rendy_shader for the reflection
- [x] All changes are feature-gated with spriv-reflection feature, enabled by default
- [x] Support for push constants
- [x] Test cases for layout merging
- [x] Example documentation

I'm looking for direction and feedback on the implementation. I know some portions require some iterating to make them more efficient (such as reflection always re-compiling and re-reflecting the shader, the entire middleware is copy and not ref, etc.) but this seems like a good baseline for basic usage of a reflector and good for expansion.

I am also not familiar enough with Vulkan to say what edge cases and further builders need to be added for this to be production ready.


.gitignore
```diff
@@ -2,3 +2,4 @@ target
 **/*.rs.bk
 **/Cargo.lock
 .DS_Store
+.idea
```
LucioFranco: you can add this to your global .gitignore

https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

jaynus: Done, I've removed the change.

rendy/examples/quads/main.rs
```diff
@@ -8,6 +8,9 @@
     allow(unused)
 )]
 
+#[cfg(not(feature = "spirv-reflection"))]
+use rendy::graph::render::SetLayout; // This is required to fix a build issue with the feature gating
```
Frizi: You should probably move the `use` into `layout` method, or just directly use fully specified path.

shader/Cargo.toml
```diff
@@ -11,10 +11,13 @@ categories = ["rendering"]
...
 gfx-hal = "0.1"
 rendy-factory = { version = "0.1", path = "../factory" }
 shaderc = { version = "0.3", optional = true }
+spirv-reflect = { version = "0.1.9", optional = true }
 
 [features]
-shader-compiler = ["shaderc"]
\ No newline at end of file
+shader-compiler = ["shaderc"]
+spirv-reflection = [ "spirv-reflect" ]
```
Frizi: I know it was like that, but could you add the trailing newline?

shader/src/lib.rs
```diff
@@ -24,6 +30,13 @@ pub trait Shader {
     /// Get spirv bytecode.
     fn spirv(&self) -> Result<std::borrow::Cow<'_, [u8]>, failure::Error>;
 
+    #[cfg(feature = "spirv-reflection")]
+    /// Uses spirv-reflect to generate a [SpirvShaderDescription] reflection representation, which is
+    /// an intermediate to gfx_hal daa representations.
```
Frizi: typo

graph/src/reflect.rs
```diff
@@ -0,0 +1,41 @@
...
+        }
+    }
+
+    fn attributes(&self) -> (Vec<gfx_hal::pso::Element<gfx_hal::format::Format>>, gfx_hal::pso::ElemStride)
+    {
+        let stride: u32 = 0;
+        let elements: Vec<gfx_hal::pso::Element<gfx_hal::format::Format>> = self.input_variables.iter()
+            .filter(|(k, _)|{
+                if k.contains("gl_") || k.is_empty() {
+                    return false
```
Frizi: Why are `gl_*` filtered out? Spirv doesn't have to be GLSL specific, and those attribute names aren't really special in any way. Spirv uses `OpDecorate` to denote their special purpose instead.

shader/src/shaderc.rs
```diff
@@ -27,14 +34,8 @@ impl<P, E> SourceShaderInfo<P, E> {
...
     }
-}
 
-impl<P, E> Shader for SourceShaderInfo<P, E>
-where
-    P: AsRef<std::path::Path> + std::fmt::Debug,
-    E: AsRef<str>,
-{
-    fn spirv(&self) -> Result<std::borrow::Cow<'static, [u8]>, failure::Error> {
+    fn compile(&self, debug: bool) -> Result<std::borrow::Cow<'static, [u8]>, failure::Error> {
```
Frizi: This isn't really `compile`. This name only makes sense for this specific impl, because here we are indeed "compiling" higher order language into spirv bytecode. Other shader implementations likely won't do any compilation in there, but just output spirv bytecode one way or another.

That way, `debug` setting should also be a part of `SourceShaderInfo` struct.

Also because you are doing analysis of this bytecode, you need to have a separate typed context for it to make it reasonable and avoid processing it every time.

What we could do is to introduce a newtype struct

```rust
struct Spirv<'a> { bytecode: Cow<'a, [u8]>> }
```
and return it from `spirv` method.

Then you can implement your reflection traits for `Spirv` type directly.

graph/src/reflect.rs
```diff
@@ -0,0 +1,41 @@
+/// Reflection extensions
+
+use rendy_shader::reflect::SpirvShaderDescription;
+use crate::node::render::{Layout, SetLayout};
+
+/// Extension for SpirvShaderReflection providing graph render type conversion
+pub trait ShaderReflectBuilder {
```
Frizi: This doesn't look like a builder type. I'm not exactly sure what you are trying to do here with this name.

shader/src/reflect.rs
```diff
@@ -0,0 +1,312 @@
...
+    bits
+}
+
+/// Implementation of shader reflection for SPIRV
+#[derive(Clone)]
+pub struct SpirvShaderDescription {
+    /// Hashmap of output variables with names.
+    pub output_variables: HashMap<String, gfx_hal::pso::AttributeDesc>,
+    /// Hashmap of output variables with names.
+    pub input_variables: HashMap<String, gfx_hal::pso::AttributeDesc>,
```
Frizi: I woudn't call it `variables`, even if it's original name from spirv-reflect, because this word might mean way too many things in context of shaders. Just `attributes` would IMO be much better.

shader/src/reflect.rs
```diff
@@ -0,0 +1,312 @@
...
+                        .filter(|v| {
+                            let r = v.reflect_into();
+                            match strict {
+                                true => { r.is_ok() && !r.as_ref().unwrap().0.is_empty() },
+                                false => r.is_ok(),
+                            }
+                        })
+                        .map(|v| {
+                            v.reflect_into().unwrap()
+                        })
```
Frizi: You should be able to replace that with `.map(ReflectInto::reflect_into)?`, as `Result`s can be propagated out from vectors and iterators.

shader/src/reflect.rs
```diff
@@ -0,0 +1,312 @@
...
+                            }
+                        })
+                        .map(|v| {
+                            v.reflect_into().unwrap()
+                        }).
+                        collect(),
+                    descriptor_sets: module.enumerate_descriptor_sets(None).map_err(|_| failure::format_err!("Cant get descriptor sets") )?.iter()
+                        .filter(|v| {
+                            match strict {
+                                true => { v.reflect_into().unwrap(); true },
```
Frizi: Obviously something is going wrong here with `strict`. You can probably make it better with this `Result` propagation trick again. 

Apart from that, what's the purpose of this flag at all? Shoudn't it always be strict?

shader/src/reflect.rs
```diff
@@ -0,0 +1,312 @@
...
+            }
+        }
+    }
+
+    ///
+    pub fn from_file<P>(path: P, strict: bool) -> Result<Self, failure::Error>
+        where
+            P: AsRef<std::path::Path> + std::fmt::Debug,
+    {
+        let mut file = File::open(path)?;
```
Frizi: This probably shoudn't be here at all. We have file loading taken care of on `Shader` trait implementations end, now we just need a way to apply reflection to those loaded shaders.

shader/src/reflect.rs
```diff
@@ -0,0 +1,312 @@
...
+        let mut contents: Vec<u8> = Vec::with_capacity(file.metadata()?.len() as usize);
+        file.read_to_end(&mut contents)?;
+
+        Self::from_bytes(contents.as_slice(), strict)
+    }
+}
+
+impl std::fmt::Debug for SpirvShaderDescription {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        for input in &self.input_variables {
```
Frizi: This manual `Debug` impl is lacking. You should use [`DebugStruct`](https://doc.rust-lang.org/std/fmt/struct.DebugStruct.html) to make it sane. Also note that `Debug` uses various flags to control the verbosity of it's input. For example, `{:#?}` will make your debug be pretty printed. Debug output shouldn't really contain newlines in default formatting.

In fact, you may want to just `#[derive(Debug)]` here and pretty-printed `{:#}` should probably do what you wanted to achieve here ;)

Frizi:
One more thing: single shader along most often isn't enough to describe a complete layout. Usually we have a vertex + fragment shader set. To make layout conversion useful, we also need a way to merge two or more reflections into a single layout.

shader/src/shaderc.rs
jaynus: Theres a problem with this - the reflected shader info is basically useless unless it is specifically compiled in debug mode. If its not, all names, debug symbols, etc. are optimized out and we have no information about the shader to actually reflect it. We can guess, but thats obviously not good enough :)

So there are two options there:
1. The SourceShaderInfo needs to implement spirv and spirv_debug, allowing consumers to provide it appropriately
2. we compile as shown here.

shader/src/reflect.rs
jaynus: I implemented the strict because I wanted in the case of "We can actually reflect some values, probably because they are not implemented, but we still want to return the basics we can't reflect"

Right now, any unsupported reflection cases for anything this supports will propagate an error. As I was developing, obviously i wanted to get past the errors for unsupported cases. While doing that, I realized it may be useful for this case - it prevents us from having to support all reflection cases, but allows us to extract the appropriate information needed for the baseline reflection. As an example, in this implementation, really the only thing supported by rendy proper is layout reflection.

shader/src/reflect.rs
jaynus: See the previous discussion about debug compilations before i make these changes.

shader/src/reflect.rs
Frizi: This is very unfortunate, and I wonder how much we should care about that. We can't realy on a shader being always compiled in rendy, because it's expected that actual shipping game will just prefer to ship compiled spirv. I would need to do some research on that. Anyway, I'd say that we should just return the whatever limited reflection is provided from that compiled spirv. I expect that names might not match, but at least the layout should be intact (or reduced to what's actually used by the shader, which is still fine).

shader/src/reflect.rs
Frizi: This still holds. If anything, we should just enable debug builds on shaders or configure it in a way to preserve some debug data like binding names. Again, I need some more research on that, but I still feel like we shouldn't be handling FS operations in reflection code.

shader/src/reflect.rs
jaynus: Yes, the layout is intact - just no names are provided which means discerning some variable purposes is impossible.

I think it will still be workable, just limited usability. 

For expanded usability, I have a few ideas in my humble and limited opinion. Shader reflection could become a build library, which reflects a separate debug shader at compile time and provides the information in a build-stage intermediate to the runtime code. This would mean shaders could not be hot-swapped without a recompile. This could be release feature gated. 

For now, I will rip out all the compilation requirements and add the additional variable handling needed for release shader processing.

Frizi:
bors retry

shader/src/lib.rs
```diff
@@ -24,6 +30,15 @@ pub trait Shader {
     /// Get spirv bytecode.
     fn spirv(&self) -> Result<std::borrow::Cow<'_, [u8]>, failure::Error>;
 
+    #[cfg(feature = "spirv-reflection")]
+    /// Uses spirv-reflect to generate a [SpirvShaderDescription] reflection representation, which is
+    /// an intermediate to gfx_hal data representations.
+    fn reflect(&self) -> Result<SpirvShaderDescription, failure::Error> {
+        Err(failure::format_err!(
+            "Reflection unimplemented for shader type"
```
Frizi: This could default to using `self.spirv()` instead of error.

graph/src/lib.rs
```diff
@@ -35,3 +35,6 @@ mod graph;
 mod node;
 
 pub use self::{graph::*, node::*};
+
+#[cfg(feature = "spirv-reflection")]
+pub mod reflect;
```
Frizi: Please add trailing newline

shader/Cargo.toml
```diff
@@ -11,13 +11,16 @@ categories = ["rendering"]
...
 shaderc = { version = "0.3", optional = true }
 serde = { version = "1.0", optional = true, features = ["derive"] }
 serde_bytes = { version = "0.10.5", optional = true }
+spirv-reflect = { git = "https://github.com/jaynus/spirv-reflect-rs.git", optional = true }
 
 [features]
 shader-compiler = ["shaderc"]
-serde-1 = ["serde", "serde_bytes"]
+spirv-reflection = [ "spirv-reflect" ]
+serde-1 = ["serde", "serde_bytes", "gfx-hal/serde" ]
```
Frizi: Add trailing newline

shader/src/lib.rs
```diff
@@ -6,6 +6,12 @@ mod shaderc;
 #[cfg(feature = "shader-compiler")]
 pub use self::shaderc::*;
 
+#[cfg(feature = "spirv-reflection")]
+pub mod reflect;
```
Frizi: group `mod`s and `use`s together.

shader/src/reflect.rs
```diff
@@ -0,0 +1,315 @@
+//! Using spirv-reflect-rs for reflection.
+//!
+
+use log::trace;
```
Frizi: we usually use `log` macros fully qualified. Just do `log::trace!` in place and remove this use.

shader/src/reflect.rs
```diff
@@ -0,0 +1,315 @@
...
+                    .map_err(|_| failure::format_err!("Error parsing descriptor sets"))?;
+                descriptor_sets_final.iter_mut().for_each(|v| {
+                    v.iter_mut().for_each(|(mut set)| {
+                        set.stage_flags = convert_stage(module.get_shader_stage())
+                    });
+                });
+
+                Ok(Self {
+                    input_attributes: input_attributes
+                        .map_err(|_| failure::format_err!("Error parsing input attributes"))?,
```
Frizi: I would put those `map_err`s directly after `collect`.

shader/src/reflect.rs
```diff
@@ -0,0 +1,315 @@
...
+    /// Hashmap of output variables with names.
+    pub descriptor_sets: Vec<Vec<gfx_hal::pso::DescriptorSetLayoutBinding>>,
+}
+
+impl SpirvShaderDescription {
+    ///
+    pub fn from_bytes(data: &[u8]) -> Result<Self, failure::Error> {
+        trace!("Shader reflecting into SpirvShaderDescription");
+
+        match ShaderModule::load_u8_data(data) {
```
Frizi:
```rust
let module = ShaderModule::load_u8_data(data).map_err(..)?;
```
Also the parsing error is probably meaningful, please don't just discard it. Either don't do the `map_err` at all if the error is well described, or use `with_context`.

shader/src/shaderc.rs
```diff
@@ -59,6 +62,12 @@ where
 
         Ok(std::borrow::Cow::Owned(artifact.as_binary_u8().into()))
     }
+
+    fn reflect(&self) -> Result<reflect::SpirvShaderDescription, failure::Error> {
```
Frizi: This would fail to compile without `spriv-reflection` feature enabled right now.

Anyway, it can be removed because the default implementation would do exactly the same.

graph/src/reflect.rs
```diff
@@ -0,0 +1,116 @@
...
+    } else {
+        if lhv.binding == rhv.binding {
+            return BindingEquality::SameBindingNonEqual;
+        }
+    }
+
+    return BindingEquality::NotEqual;
+}
+
+impl ShaderLayoutGenerator for (SpirvShaderDescription, SpirvShaderDescription) {
```
Frizi: I'm not a fan of this. Tuple based APIs usually end up requiring macros for multiple impls and are generally limited and harder to work with.

How about doing an iterator based API instead?

Also this method returns different type than it's input `(A, A) => B`, which makes it impossible to fold over, thus merge more than two descriptions. We would need either `(A, A) => A` or `(B, A) => B` with initial empty `B`. Would it be possible to either merge two layouts into layout or two descriptions into description?

shader/src/reflect.rs
jaynus: These map_err's are specifically for spirv-reflect calls instead of internal to reflection. I wanted to map those errors seperately, as the actual parse errors are then handles by the upper result and map error. later at line 303. I can't ceorce the types for the Results without the separation as I get type inference errors mapping them here. I may just be dumb though and not understand how to do it. I've added detail errors, though.

graph/src/reflect.rs
```diff
@@ -0,0 +1,218 @@
...
+    where T: IntoIterator,
+          T::Item: Shader + Sized
+{
+    fn merge(mut self, ) -> Result<Layout, failure::Error> {
+        let mut iter = self.into_iter();
+
+        let mut sets = Vec::new();
+
+        let mut shader = iter.next();
+        while shader.is_some() {
```
Frizi: The better syntax for that would be `for shader in iter`

graph/src/reflect.rs
```diff
@@ -0,0 +1,218 @@
...
+            for (n, set) in current_layout.sets.iter().enumerate() {
+                if n < sets.len() {
+                    // The set already exists, lets make sure we match or are a subset of it. If we are a Superset of it, we should replace it.
+                    if let Some(existing_set) = sets.get(n) {
+                        match compare_set(set, existing_set) {
+                            SetEquality::NotEqual => {
+                                return Err(failure::format_err!("Mismatching bindings between shaders for set #{}", n));
+                            },
+                            SetEquality::SupersetOf => {
+                                sets.insert(n, set.clone()); // Overwrite it
```
Frizi: I believe that insert will not replace the element, but shift to the right and insert in new space.

graph/src/reflect.rs
```diff
@@ -0,0 +1,218 @@
...
+                            },
+                            SetEquality::Equal | SetEquality::SubsetOf => {
+                                for binding in sets.get_mut(n).unwrap().bindings.iter_mut() {
```
Frizi: instead of `sets.get_mut(n).unwrap()`, this can be just `sets[n]`

graph/src/reflect.rs
```diff
@@ -0,0 +1,218 @@
...
+
+        let mut shader = iter.next();
+        while shader.is_some() {
+            let s = shader.unwrap();
+            let current_layout = s.layout()?;
+
+            for (n, set) in current_layout.sets.iter().enumerate() {
+                if n < sets.len() {
+                    // The set already exists, lets make sure we match or are a subset of it. If we are a Superset of it, we should replace it.
+                    if let Some(existing_set) = sets.get(n) {
```
Frizi: I would write this as `match sets.get(n).map(|existing| compare_set(set, existing))`. This would allow to get rid of both `if n < sets.len()` and `if let Some(existing_set) = sets.get(n)`, and handle all the cases in one match block.

```rust
match sets.get(n).map(|existing| compare_set(set, existing)) {
  None => sets.push(set.clone()),
  Some(SetEquality::NotEqual) => {...}
  ...
}
```

shader/src/reflect.rs
```diff
@@ -0,0 +1,319 @@
...
+}
+
+fn type_element_format(
+    flags: variable::ReflectTypeFlags,
+    traits: &traits::ReflectTypeDescriptionTraits,
+) -> Result<gfx_hal::format::Format, failure::Error> {
+    let mut current_type = Format::R32Float;
+
+    if flags.contains(variable::ReflectTypeFlags::INT) {
+        current_type = match traits.numeric.scalar.signedness {
```
Frizi: Those matches can be combined
```rust
match (traits.numeric.scalar.signedness, traits.numeric.scalar.width) {
    (1, 8) => Format::R8Int,
    (1, 16) => Format::R16Int,
    ...
}
```

In fact, i would go a little bit further and combine that with `INT/FLOAT`, something like that

```rust
enum NumTy {
  Int(u32),
  Float,
}

let num_ty = if flags.contains(variable::ReflectTypeFlags::INT) {
  NumTy::Int(traits.numeric.scalar.signedness)
} else if if flags.contains(variable::ReflectTypeFlags::FLOAT) {
  NumTy::Float
} else {
  failure::bail!("Unrecognized numeric type with flags {}", flags);
};

let current_type = match (num_ty, traits.numeric.scalar.width) {
    (NumTy::Int(1), 8) => Format::R8Int,
    (NumTy::Int(1), 16) => Format::R16Int,
    ...
    (NumTy::Float, 32) => Format::R32Float,
    ...
}```

File: shader/src/reflect.rs [hunk 27]
```diff
@@ -0,0 +1,319 @@
...
+        // TODO: support other bits
+        current_type = match traits.numeric.scalar.width {
+            32 => Format::R32Float,
+            64 => Format::R64Float,
+            _ => return Err(failure::format_err!("Unrecognized scalar width for float")),
+        }
+    }
+
+    if flags.contains(variable::ReflectTypeFlags::VECTOR) {
+        current_type = match traits.numeric.vector.component_count {
```
Frizi: Again easy to combine the match
```rust
match (current_type, traits.numeric.vector.component_count) {
    (Format::R64Float, 2) => Format::Rg64Float,
    ...
}

File: shader/src/reflect.rs [hunk 28]
```diff
@@ -0,0 +1,319 @@
...
+//
+
+impl ReflectInto<gfx_hal::pso::DescriptorType> for descriptor::ReflectDescriptorType {
+    fn reflect_into(&self) -> Result<gfx_hal::pso::DescriptorType, failure::Error> {
+        match *self {
+            descriptor::ReflectDescriptorType::Sampler => Ok(gfx_hal::pso::DescriptorType::Sampler),
+            descriptor::ReflectDescriptorType::CombinedImageSampler => {
+                Ok(gfx_hal::pso::DescriptorType::CombinedImageSampler)
+            }
+            descriptor::ReflectDescriptorType::SampledImage => {
```
Frizi: some local `use` statements on the beginning of this function would help declutter this
```rust
use descriptor::ReflectDescriptorType::*;
use gfx_hal::pso::DescriptorType;
```

shader/src/reflect.rs
```diff
@@ -0,0 +1,319 @@
...
+
+impl ReflectInto<Vec<gfx_hal::pso::DescriptorSetLayoutBinding>>
+    for descriptor::ReflectDescriptorSet
+{
+    fn reflect_into(
+        &self,
+    ) -> Result<Vec<gfx_hal::pso::DescriptorSetLayoutBinding>, failure::Error> {
+        let mut output = Vec::<gfx_hal::pso::DescriptorSetLayoutBinding>::new();
+
+        for descriptor in self.bindings.iter() {
```
Frizi: Loop can be rewritten as
```rust
self.bindings.iter().map(|desc| desc.reflect_into).collect()
```

shader/src/shaderc.rs
```diff
@@ -63,7 +79,9 @@ where
                         .ok_or_else(|| failure::format_err!("Failed to init Shaderc"))?;
                     ops.set_target_env(shaderc::TargetEnv::Vulkan, vk_make_version!(1, 0, 0));
                     ops.set_source_language(self.lang);
-                    ops.set_optimization_level(shaderc::OptimizationLevel::Performance);
+                    //ops.set_generate_debug_info();
+                    //ops.set_optimization_level(shaderc::OptimizationLevel::None);
```
Frizi: clean up comments and enable debug info under a feature flag
```rust
#[cfg(feature = "spirv-reflection")]
ops.set_generate_debug_info();
```

shader/src/lib.rs
```diff
@@ -62,3 +73,36 @@ impl Shader for SpirvShader {
...
     }
 }
+
+#[cfg(feature = "spirv-reflection")]
+/// Spir-V shader with reflection
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Clone, Debug)]
+pub struct SpirvReflectedShader {
+    #[cfg_attr(feature = "serde", serde(with = "serde_bytes"))]
+    spirv: Vec<u8>,
```
Frizi: we don't necessarily need to have `spirv` vec left around forever.
Instead we can directly return a `SpirvShaderDescription` struct.

then shader would have a `fn reflect(&self) -> Result<SpirvShaderDescription, failure::Error>` and that shader description would be kept inside the pass.

jaynus: That was the original design we had, where we had multiple iterations of introspection happening. I changed it to this design to prevent a few copies of the introspection, at the cost of using a different Spirv struct for the reflected case. I can change this back if need be.

graph/src/reflect.rs
```diff
@@ -0,0 +1,310 @@
...
+
+/// Iterator function implementation for merging multiple shader layouts into a single shader set layout
+pub trait SpirvLayoutMerger {
+    /// Returns a merged Layout from the provided iterator
+    fn merge(self) -> Result<Layout, failure::Error>;
+}
+impl<T> SpirvLayoutMerger for T
+where
+    T: IntoIterator,
+    T::Item: Shader + Sized,
```
Frizi: This could accept `T: IntoIterator<Item = Layout>` instead. That way it is more clear that it merges multiple layouts, and you don't have a dependency on shader type.

graph/src/reflect.rs
```diff
@@ -0,0 +1,310 @@
...
+{
+    fn merge(self) -> Result<Layout, failure::Error> {
+        let mut sets = Vec::new();
+        let mut push_constants = Vec::new();
+
+        for s in self.into_iter() {
```
Frizi: that would then turn into `for layout in self.into_iter()`

Frizi: in fact we should have merged for the whole reflection

jaynus:
Closing in favor of #103

PR #73, opened 2019-03-16
Descriptors

omni-viral:


descriptor/src/allocator.rs
```diff
@@ -0,0 +1,272 @@
...
+            let allocate = min(pool.free, count);
+            log::trace!("Allocate {} from exising pool", allocate);
+            allocate_from_pool::<B>(&mut pool.raw, layout.raw(), allocate, &mut allocation.sets)?;
+            allocation.pools.extend(
+                std::iter::repeat(index as u64 + self.pools_offset).take(allocate as usize),
+            );
+            count -= allocate;
+            pool.free -= allocate;
+
+            if count == 0 {
```
Frizi: This test could also be done on the beginning to avoid iterating over pools and printing stray logs. I see it's done outside of that method, but it's easy to forget about it. Compiler should prove it unnecessary and get rid of it anyway.

descriptor/src/allocator.rs
```diff
@@ -0,0 +1,272 @@
...
+            }
+            Err(err) => {
+                // Free sets allocated so far.
+                let mut last = None;
+                for (index, pool) in self.allocation.pools.drain(..).enumerate().rev() {
+                    match last {
+                        Some(last) if last != pool => {
+                            bucket.free(self.allocation.sets.drain(index + 1..), last);
+                        }
+                        _ => last = Some(pool),
```
Frizi: I don't quite understand that logic. Why a single allocation error triggers freeing everything but last pool?

descriptor/src/allocator.rs
```diff
@@ -0,0 +1,272 @@
...
+                Some((ranges, pool, sets)) => {
+                    self.buckets
+                        .get_mut(&ranges)
+                        .expect("Set should be allocated from this allocator")
+                        .free(sets.drain(), *pool);
+                    *pool = set.pool;
+                    sets.push(set.raw);
+                }
+                slot @ None => {
+                    *slot = Some((set.ranges, set.pool, smallvec![set.raw]));
```
Frizi: could as well use `.replace` on the slot option for some extra readability points.

descriptor/src/allocator.rs
```diff
@@ -0,0 +1,272 @@
...
+            }
+        }
+    }
+
+    pub unsafe fn free(&mut self, sets: impl IntoIterator<Item = DescriptorSet<B>>) {
+        let mut free: Option<(DescriptorRanges, u64, SmallVec<[B::DescriptorSet; 32]>)> = None;
+        for set in sets {
+            match &mut free {
+                Some((ranges, pool, sets)) if *ranges == set.ranges && *pool == set.pool => {
+                    sets.push(set.raw);
```
Frizi: This name shadowing got me lost for a moment. Please change the `sets` names to unique ones.

descriptor/src/allocator.rs
```diff
@@ -0,0 +1,272 @@
...
+                    }
+                }
+                Err(err)
+            }
+        }
+    }
+
+    pub unsafe fn free(&mut self, sets: impl IntoIterator<Item = DescriptorSet<B>>) {
+        let mut free: Option<(DescriptorRanges, u64, SmallVec<[B::DescriptorSet; 32]>)> = None;
+        for set in sets {
```
Frizi: I'm having hard time with understanding that control flow. Clearly the `None` pattern is hit first every time, so it could as well be the first one in code.
Alas, i still don't really understand what is this doing. Some explanation would be very helpful.

descriptor/src/ranges.rs
```diff
@@ -0,0 +1,171 @@
...
+];
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub struct DescriptorRanges {
+    counts: [u32; DESCPTOR_TYPES_COUNT],
+}
+
+impl DescriptorRanges {
+    pub fn zero() -> Self {
+        DescriptorRanges { counts: [0; 11] }
```
Frizi: `11` -> `DESCPTOR_TYPES_COUNT`

descriptor/src/ranges.rs
```diff
@@ -0,0 +1,171 @@
...
+        };
+
+        for binding in bindings {
+            descs.counts[binding.ty as usize] += binding.count as u32;
+        }
+
+        descs
+    }
+
+    pub fn promote_to_power_of_two(&self) -> Self {
```
Frizi: What's the reason behind that method? When should I need to inflate my layout size?

factory/src/factory.rs
```diff
@@ -544,6 +624,75 @@ where
...
+
+    /// Create descriptor set layout with specified bindings.
+    pub fn create_descriptor_set_layout(
+        &self,
+        bindings: Vec<DescriptorSetLayoutBinding>,
+    ) -> Result<DescriptorSetLayout<B>, OutOfMemory> {
+        DescriptorSetLayout::create(&self.device, bindings)
+    }
+
+    /// Create descriptor set layout with specified bindings.
```
Frizi: Certainly not "Create" :D

wsi/src/lib.rs
```diff
@@ -106,7 +108,7 @@ pub struct Surface<B: gfx_hal::Backend> {
     window: std::sync::Arc<winit::Window>,
     raw: B::Surface,
     #[cfg(not(feature = "no-slow-safety-checks"))]
```
Frizi: As `RendySlowId` is now ZST when feature is disabled, this `cfg` can go away. Compiler will remove all related code anyway.

Also noticed that `instance_id()` is now unguarded, so this should now trigger a compile error when feature is enabled.

wsi/src/lib.rs
```diff
@@ -128,11 +130,15 @@ where
...
     ) -> Self {
         let raw = create_surface::<B>(instance, &window);
         #[cfg(not(feature = "no-slow-safety-checks"))]
-        let s = Surface { window, raw, factory_id: _factory_id };
+        let s = Surface {
+            window,
+            raw,
+            instance_id,
+        };
         #[cfg(feature = "no-slow-safety-checks")]
```
Frizi: This should also go away

descriptor/src/allocator.rs
omni-viral: Only sets that was allocated during current allocations are freed.
And I forgot to free last one.

descriptor/src/ranges.rs
omni-viral: On the second thought you never want it. I'll remove this method.

factory/src/factory.rs
omni-viral: Copypasta 😅

descriptor/src/allocator.rs
Frizi: All good now 👍

omni-viral:
Bors r=Frizi

termhn:
bors retry

PR #75, opened 2019-03-19
Make relevant backtraces feature gated

termhn:
They by default pollute panic backtraces terribly so you have to scroll past a load of spam to get to the actual panic backtrace

### DIFF
--- a/graph/Cargo.toml
+++ b/graph/Cargo.toml
@@ -15,7 +15,7 @@ empty = ["gfx-backend-empty", "rendy-wsi/gfx-backend-empty", "rendy-factory/gfx-
 dx12 = ["gfx-backend-dx12", "rendy-wsi/gfx-backend-dx12", "rendy-factory/gfx-backend-dx12"]
 metal = ["gfx-backend-metal", "rendy-wsi/gfx-backend-metal", "rendy-factory/gfx-backend-metal"]
 vulkan = ["gfx-backend-vulkan", "rendy-wsi/gfx-backend-vulkan", "rendy-factory/gfx-backend-vulkan"]
-relevant-backtrace = ["rendy-factory/relevant-backtrace", "rendy-command/relevant-backtrace", "rendy-wsi/relevant-backtrace", "rendy-frame/relevant-backtrace", "rendy-memory/relevant-backtrace", "rendy-resource/relevant-backtrace", "relevant/backtrace"]
+relevant-backtrace = ["relevant/backtrace"]
 
 [dependencies]
 rendy-chain = { version = "0.1.0", path = "../chain" }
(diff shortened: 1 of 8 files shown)


graph/Cargo.toml
```diff
@@ -15,6 +15,7 @@ empty = ["gfx-backend-empty", "rendy-wsi/gfx-backend-empty", "rendy-factory/gfx-
 dx12 = ["gfx-backend-dx12", "rendy-wsi/gfx-backend-dx12", "rendy-factory/gfx-backend-dx12"]
 metal = ["gfx-backend-metal", "rendy-wsi/gfx-backend-metal", "rendy-factory/gfx-backend-metal"]
 vulkan = ["gfx-backend-vulkan", "rendy-wsi/gfx-backend-vulkan", "rendy-factory/gfx-backend-vulkan"]
+relevant-backtrace = ["rendy-factory/relevant-backtrace", "rendy-command/relevant-backtrace", "rendy-wsi/relevant-backtrace", "rendy-frame/relevant-backtrace", "rendy-memory/relevant-backtrace", "rendy-resource/relevant-backtrace", "relevant/backtrace"]
```
omni-viral: It is enough to specify just `relevant-backtrace = ["relevant/backtrace"]` in crates that depend on `relevant`.
No need in enabling feature in all dependencies as the only this it should do is enable feature in `relevant` crate.

rendy/Cargo.toml
```diff
@@ -32,6 +32,8 @@ texture = ["rendy-texture", "factory", "util"]
 util = ["rendy-util"]
 wsi = ["rendy-wsi"]
 
+relevant-backtrace = ["rendy-factory/relevant-backtrace", "rendy-graph/relevant-backtrace", "rendy-wsi/relevant-backtrace", "rendy-memory/relevant-backtrace", "rendy-resource/relevant-backtrace", "rendy-command/relevant-backtrace", "rendy-frame/relevant-backtrace"]
```
omni-viral: Better specify just enable optional dependency on `relevant` and make it `relevant-backtrace = ["relevant/backtrace"]`. Otherwise this feature became unusable for people who don't enable all subcrates.

termhn: I'm not sure how this would work -- if someone isn't using all subcrates then they'd need to make their own relevant-backtrace feature for all the subcrates anyway, no?

omni-viral: You can specify feature once anywhere and crate will have, all dependent crates observe feature enabled.

memory/Cargo.toml
```diff
@@ -12,6 +12,7 @@ description = "Rendy's memory manager"
 
 [features]
 serde-1 = ["serde"]
+relevant-backtrace = ["relevant/backtrace"]
```
omni-viral: You didn't remove `backtrace` feature from dependency below.

omni-viral:
BTW. I made it so relevant types stay silent if they dropped due to panic.

termhn:
I think this can be closed then

PR #81, opened 2019-03-23
Escape descriptor sets from factory

omni-viral:


descriptor/src/allocator.rs
```diff
@@ -132,6 +133,30 @@ where
             return Ok(());
         }
 
+        for (index, pool) in self.pools.iter_mut().enumerate().rev() {
```
omni-viral: You can't reuse set from pool bc it may be allocated for different layout (with same descriptors count, but still)

descriptor/src/allocator.rs
```diff
@@ -166,19 +166,24 @@ where
                 pool_ranges,
             );
             let raw = device.create_descriptor_pool(size as usize, &pool_ranges)?;
+            let allocate = size.min(count);
 
             self.pools.push_back(DescriptorPool {
                 raw,
                 size,
-                free: size,
+                free: size - allocate,
```
omni-viral: What if allocation will fail?

resource/src/resources.rs
```diff
@@ -363,6 +425,21 @@ where
             block.map(|block| heaps.free(device, block));
         }
 
+        descriptor_allocator.free(
```
omni-viral: Sets on terminal shoild be dropped too.

omni-viral:
LGTM!

Frizi:
bors retry

omni-viral:
bors retry

PR #82, opened 2019-03-25
Fix allocator disposal error message

Frizi:
`self.pools` doesn't contain the pool that triggered the error, and errors are repeated for each pool anyway.


omni-viral:
bors retry

PR #83, opened 2019-03-27
Add customizable input assembly desc

fhaynes:
Adds support for allowing specification of the input assembly desc defaulting to TriangleList with no restart but overrideable.


fhaynes:
Also please note I have no idea what I'm doing.

PR #88, opened 2019-04-03
Small resource system refactoring

omni-viral:


memory/src/mapping/mod.rs
```diff
@@ -43,44 +43,47 @@ impl<'a, B> MappedRange<'a, B>
...
-            fits_usize(range.end - range.start),
-            "Range length must fit in usize"
-        );
-        assert!(memory.host_visible());
+    // /// Map range of memory.
+    // /// `range` is in memory object space.
+    // ///
+    // /// # Safety
+    // ///
+    // /// * Only one range for the given memory object can be mapped.
```
termhn: Delete instead of comment?

memory/src/mapping/mod.rs
```diff
@@ -122,15 +126,26 @@ where
         'a: 'b,
         T: Copy,
     {
+        assert!(
```
termhn: return err instead of panic?

memory/src/mapping/mod.rs
```diff
@@ -122,15 +126,26 @@ where
...
+            range.start < range.end,
+            "Memory mapping region must have valid size"
+        );
+        assert!(
```
termhn: same here

memory/src/mapping/mod.rs
```diff
@@ -149,15 +164,26 @@ where
         'a: 'b,
         T: Copy,
     {
+        assert!(
```
termhn: err instead of panic again

resource/src/escape.rs
```diff
@@ -15,84 +14,36 @@ use {
...
-            forget(self);
-
-            match Arc::try_unwrap(inner) {
-                Ok(_) => Some(value),
-                Err(inner) => {
-                    inner.escape(value);
-                    None
-                }
-            }
+            std::ptr::read(&mut escape.sender);
```
termhn: why read without doing anything with it?

resource/src/resources.rs
```diff
@@ -19,474 +11,79 @@ pub struct Epochs {
...
 
-    /// Wrap a descriptor_set.
-    pub fn create_descriptor_sets<E>(
-        &self,
-        device: &impl gfx_hal::Device<B>,
-        allocator: &mut DescriptorAllocator<B>,
-        layout: &DescriptorSetLayout<B>,
-        count: u32,
-    ) -> Result<E, gfx_hal::device::OutOfMemory>
+    /// Wrap resource instance into handle.
```
termhn: into Escape

resource/src/resources.rs
```diff
@@ -19,474 +11,79 @@ pub struct Epochs {
...
-        device: &impl gfx_hal::Device<B>,
-        bindings: Vec<gfx_hal::pso::DescriptorSetLayoutBinding>,
-    ) -> Result<DescriptorSetLayout<B>, gfx_hal::device::OutOfMemory> {
-        Ok(unsafe {
-            DescriptorSetLayout::new(
-                descriptor::DescriptorSetLayout::create(device, bindings)?,
-                &self.descriptor_set_layouts,
-            )
-        })
+    /// Wrap resource instance into handle.
```
termhn: into Handle

resource/src/resources.rs
```diff
@@ -19,474 +11,79 @@ pub struct Epochs {
...
-    descriptor_sets: Terminal<(descriptor::DescriptorSet<B>, KeepAlive)>,
-    descriptor_set_layouts: Terminal<descriptor::DescriptorSetLayout<B>>,
-    sampler_cache: SamplerCache<B>,
-
-    dropped_buffers: VecDeque<(Epochs, B::Buffer, MemoryBlock<B>)>,
-    dropped_images: VecDeque<(Epochs, B::Image, Option<MemoryBlock<B>>)>,
-    dropped_image_views: VecDeque<(Epochs, B::ImageView, KeepAlive)>,
-    dropped_descriptor_sets: VecDeque<(Epochs, descriptor::DescriptorSet<B>, KeepAlive)>,
-    dropped_descriptor_set_layouts: VecDeque<(Epochs, descriptor::DescriptorSetLayout<B>)>,
+pub struct Resources<T> {
```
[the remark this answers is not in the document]
omni-viral: Good idea!

factory/src/factory.rs
```diff
@@ -206,14 +282,20 @@ where
...
-        buffer: &mut Buffer<B>,
+        buffer: &Buffer<B>,
         offset: u64,
         content: &[T],
         last: Option<BufferState>,
         next: BufferState,
     ) -> Result<(), failure::Error> {
         let content_size = content.len() as u64 * std::mem::size_of::<T>() as u64;
-        let mut staging = self.create_buffer(256, content_size, buffer::UploadBuffer)?;
+        let mut staging = self.create_buffer(
```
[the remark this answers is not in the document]
omni-viral: We could. But linear allocator is O(1) so this shouldn't be a problem.

memory/src/mapping/mod.rs
omni-viral: No. This is program fault. It provides clearly invalid arguments.
Probably even out of bounds and invalid access errors should result in panic.

resource/src/escape.rs
omni-viral: To drop :)
I probably should use `ManuallyDrop::drop` to make intent clear.

factory/src/factory.rs
```diff
@@ -158,7 +158,12 @@ where
         Ok(())
     }
 
-    /// Creates a buffer that is managed with the specified properties.
+    /// Creates a buffer with the specified properties.
+    ///
+    /// This function returns relevant value, that it, the value cannot be dropped.
```
termhn: 👍 looks good, one more small thing, should be "that is" rather than "that it" :)

omni-viral: :facepalm:

resource/src/escape.rs
omni-viral: I guess I should use `drop_in_place` instead.

omni-viral: And reorder functions, just in case `<Sender as Drop>::drop` will panic.

command/src/family/mod.rs
```diff
@@ -3,48 +3,62 @@
...
-        self.0
-    }
+    /// Device id.
+    pub device: DeviceId,
+}
 
-    /// Get index of the queue.
-    pub fn index(&self) -> usize {
-        self.1
+impl FamilyId {
```
termhn: `impl From<FamilyId> for gfx_hal::queue::QueueFamilyId` instead?

command/src/family/mod.rs
```diff
@@ -168,23 +177,37 @@ where
...
 {
     /// Get queue family by id.
     pub fn family(&self, id: FamilyId) -> &Family<B> {
-        &self.families[self.families_indices[id.0]]
+        assert_eq!(id.device, self.device);
+        &self.families[self.families_indices[id.index]]
+    }
+
+    /// Get queue family by index.
+    pub fn family_by_index(&self, index: usize) -> &Family<B> {
```
termhn: Unsafe since we aren't checking device is correct? Also, use this in `Families::family()` to avoid code duplication

command/src/family/mod.rs
```diff
@@ -168,23 +177,37 @@ where
...
     }
 
     /// Get queue family by id.
     pub fn family_mut(&mut self, id: FamilyId) -> &mut Family<B> {
-        &mut self.families[self.families_indices[id.0]]
+        assert_eq!(id.device, self.device);
+        &mut self.families[self.families_indices[id.index]]
+    }
+
+    /// Get queue family by index.
```
termhn: Same here

command/src/lib.rs
```diff
@@ -10,6 +10,75 @@
...
+            B: gfx_hal::Backend,
+        {
+            /// Get owner id.
+            pub fn family_id(&self) -> $crate::FamilyId {
+                ($getter)(self)
+            }
+
+            /// Assert specified family is owner.
+            pub fn assert_family_owner(&self, family: &$crate::Family<B, C>) {
+                assert_eq!(self.family_id(), family.id(), "Resource is not owned by speicified family");
```
termhn: speicified -> specified

descriptor/src/allocator.rs
```diff
@@ -117,7 +117,7 @@ where
             .min(MAX_SETS) // capped to MAX_SETS
     }
 
-    unsafe fn dispose(mut self, device: &impl Device<B>) {
+    unsafe fn dispose(mut self, device: &B::Device) {
```
termhn: Here things are going from `&impl Device<B>` -> `&B::Device` while in `factory.rs` they are going from `&B::Device` to `&Device<B>`, should probably choose one and stick to it

factory/src/factory.rs
```diff
@@ -625,6 +715,10 @@ where
     }
 
     /// Destroy fence.
+    ///
+    /// # Safety
+    ///
+    /// Fence must be created by this `Factory`.
     pub fn destroy_fence(&self, fence: Fence<B>) {
```
termhn: `assert_device_owned` here?

factory/src/factory.rs
```diff
@@ -702,7 +800,7 @@ where
         &self,
         bindings: Vec<DescriptorSetLayoutBinding>,
     ) -> Result<DescriptorSetLayout<B>, OutOfMemory> {
-        unsafe { DescriptorSetLayout::create(&self.device, set::Info { bindings }) }
+        unsafe { DescriptorSetLayout::create(&self.device, DescriptroSetInfo { bindings }) }
```
termhn: I'm too lazy to go find where the declaration of this is, but `DescriptroSetInfo` -> `DescriptorSetInfo`

resource/src/set.rs
```diff
@@ -1,82 +1,100 @@
...
     gfx_hal::{pso::DescriptorSetLayoutBinding, Backend, Device as _},
     relevant::Relevant,
     smallvec::SmallVec,
 };
 
 /// Descriptor set layout info.
 #[derive(Clone, Debug)]
-pub struct Info {
-    /// Bindings
+pub struct DescriptroSetInfo {
```
termhn: Ah, here we go. `DescriptroSetInfo` -> `DescriptorSetInfo`

util/src/wrap.rs
```diff
@@ -0,0 +1,204 @@
+use {
+    gfx_hal::Backend,
+    std::{any::Any, marker::PhantomData, ops::Deref},
+};
+
+#[cfg(not(feature = "no-slow-safety-checks"))]
+fn new_instance_id() -> InstanceId {
+    static INSTANCE_ID: std::sync::atomic::AtomicUsize = std::sync::atomic::AtomicUsize::new(0);
```
termhn: Shouldn't these be declared outside the function?

util/src/wrap.rs
```diff
@@ -0,0 +1,204 @@
...
+
+    DeviceId {
+        id: id as u32,
+        instance,
+    }
+}
+
+#[cfg(feature = "no-slow-safety-checks")]
+fn new_instance_id() -> InstanceId {
+    InstanceId {}
```
termhn: TODO here or make it actually make a new instance id

Frizi: I think this is intentional. When safety checks are off, all the instances are ZST and optimized out.

util/src/wrap.rs
```diff
@@ -0,0 +1,204 @@
+use {
+    gfx_hal::Backend,
+    std::{any::Any, marker::PhantomData, ops::Deref},
+};
+
+#[cfg(not(feature = "no-slow-safety-checks"))]
+fn new_instance_id() -> InstanceId {
+    static INSTANCE_ID: std::sync::atomic::AtomicUsize = std::sync::atomic::AtomicUsize::new(0);
+
+    log::warn!("Slow safety checks are enabled! Disable them in production by enabling the 'no-slow-safety-checks' feature!");
```
Frizi: I think this warn should be limited to only single execution, like when `fetch_add` output is 1.

util/src/wrap.rs
termhn: 🤦‍♀️ of course, somehow I didn't see this was the non-safety check version

descriptor/src/allocator.rs
omni-viral: memory and descriptor crate don't use it on purpose.
They are usable on their own. Requiring wrapped device will make it harder to use outside rendy.

factory/src/factory.rs
omni-viral: Fence will check that

util/src/wrap.rs
omni-viral: Only if it used somewhere else.

util/src/wrap.rs
omni-viral: Probably

command/src/family/mod.rs
omni-viral: No. Why? It's just vector lookup. If you'll get `Family` of wrong device you will hit later when you will try to use it.

PR #92, opened 2019-04-06
Added documentation and ran cargo fmt

fhaynes:


texture/src/pixel.rs
```diff
@@ -48,7 +50,7 @@ pub struct _32;
 #[derive(Clone, Copy, Debug, Default)]
 pub struct _64;
 
-/// Size of the channel.
+/// Size of the R, G, B, and A channels
```
termhn: Not every format has an RGB and A channel... more accurately this would be "Byte size of each channel in the image"

texture/src/format/palette.rs
```diff
@@ -1,3 +1,7 @@
+//! Module that generates functions to load a single color into a 1x1 `Texture`. Which function
+//! to use depends on what color space the user is using.
+//!
+//! TODO: Consider renaming this as there are no palettes to be found
```
termhn: Name is becuase it's designed to integrate with the `palette` crate

texture/src/pixel.rs
```diff
@@ -216,7 +225,11 @@ where
     pub repr: <C as PixelRepr<S, T>>::Repr,
 }
 
-/// Pixel trait.
+/// AsPixel trait for converting specific things into Pixels.
```
termhn: "For extracting underlying data representation information from a Rust data type"

texture/src/texture.rs
```diff
@@ -209,6 +211,7 @@ impl<'a> TextureBuilder<'a> {
             )?
             .into();
 
+        // TODO: @omni-viral can you replace this with a comment explaining the use of unsafe here?
```
termhn: The reason that `factory.upload_image` is unsafe is that the image being uploaded must have been created by the same factory and that it is not in use; we guarantee that here because we just created the image on the same factory right before.

fhaynes:
> Thanks! Just a few things :)

Thanks! Fixed, I think.

omni-viral:
Nightly has new rule for doc comments.
Could you fix it please?

fhaynes:
> Nightly has new rule for doc comments.
> Could you fix it please?

Fixed!

PR #93, opened 2019-04-07
Adds a doc briefly summarizing pipelines

fhaynes:
I am writing these as I document rendy and tinker with it and Vulkan. I thought it might be useful to include in the docs folder; feel free to close if it isn't what you are wanting in there.

(And if any of it is wrong, please _please_ correct me)

### DIFF
--- a/docs/pipelines.md
+++ b/docs/pipelines.md
@@ -128,6 +128,8 @@ Once you are ready, you can check out this full diagram of a Pipeline: https://v
 
 And then, once you've recovered, the entire chapter on Pipelines in the Vulkan spec is here: https://vulkan.lunarg.com/doc/view/1.0.33.0/linux/vkspec.chunked/ch09.html.
 
+A similar document to this by @termhn (one of the primary Rendy contributors) can be found [here](https://github.com/termhn/gfx-hal-tutorial/blob/master/articles/zero-to-voxel-render-part1.md#the-graphics-pipeline).
+
 Good Luck....
 
 _...the final piece of the note appears to be stained with a mixture of tears and something darker, possibly blood..._
--- a/docs/pipelines.md
+++ b/docs/pipelines.md
@@ -1,0 +1,131 @@
+# Graphics Pipelines
+
+This document covers the key points of a Graphics Pipeline as used by rendy and gfx-hal; they are essentially the same as a Vulkan graphics pipeline.
+
+*Please note this is not intended to be a tutorial on Vulkan. It covers only enough to understand code written in Rendy*
+
+## Pipeline Overview
+
+When you want to draw a 3D shape on someone's monitor, say a cube, you are really just drawing some number of triangles.
+
+## A Cube
+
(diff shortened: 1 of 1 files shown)


docs/pipelines.md
```diff
@@ -0,0 +1,131 @@
...
+> grid in framebuffer coordinates are occupied by the primitive, and 
+> assigning one or more depth values to each such square.
+
+Now we're making progress! Only 28,233 more stages to go!
+
+### Fragment Shader
+
+Yes, this is a shader stage that isn't with the rest of the shader stages. Shocking.
+
+As something is rasterized, the rasterizer produces a series of framebuffer addresses and values using a two-dimensional description of a point, line segment, or triangle. These are called _fragments_.
```
termhn: These fragments are analogous to pixels, however they are distingued as fragments because they have yet to undergo blending. Fragments represent a "possible incoming color" at a certain position on the frame buffer, however the final pixel color could change based on the blend stage.

docs/pipelines.md
```diff
@@ -0,0 +1,131 @@
...
+
+You probably guessed that a fragment shader alters these fragments.
+
+Once they are done, the fragments are updated in the framebuffer.
+
+### Framebuffer Operations
+
+After the fragment shaders are done, other things read the updated data to come up with the final color, transparency, and all that which we will skip. You don't have to do any of that. Fragment shaders are the only way to affect those.
```
termhn: You can affect how blending works by altering blend operation in the pipeline

docs/pipelines.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Back to Rendy
+
+In Rendy, we describe one of these pipelines using two `Trait`s: a Pipeline Descriptor, and a Pipeline. 
+
+### SimpleGraphicsPipelineDesc
+
+This `Trait` is defined in `graph/src/node/render/group/simple.rs`, and you will use it to set up the resources your pipeline will use. The `Desc` is short for `Descriptor`.
```
termhn: Desc is short for Description* which is an important distinction as descriptor has a specific meaning in regards to Vulkan/gfx-hal

fhaynes: > Desc is short for Description* which is an important distinction as descriptor has a specific meaning in regards to Vulkan/gfx-hal

Doh, thanks. I had been reading about Descriptors, didn't even notice.

docs/pipelines.md
fhaynes: I confess that this part:
```
As something is rasterized, the rasterizer produces a series of framebuffer addresses and values using a two-dimensional description of a point, line segment, or triangle.
```
is cut and pasted from the API docs. I'll add more for context..

docs/pipelines.md
fhaynes: Ooh, didn't know that. Added.

docs/pipelines.md
fhaynes: @termhn is the new version better?

PR #94, opened 2019-04-08
Universal palette conversion

omni-viral:
Add implementations to convert from whatever standard and component for rgb/rgba and luma/lumaa types.


Frizi:
bors retry

PR #95, opened 2019-04-09
Added a brief document describing memory usage in rendy and vulkan

fhaynes:


docs/buffers.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Memory Visibility and Coherency
+
+Can the CPU see the contents of the GPU's memory? Can the GPU access the system RAM? Good questions! The answer is, it depends. A graphics device will have capabilities, and it is up to you, intrepid programmer, to figure out what those are and code accordingly.
+
+A piece of memory is referred to as a `heap`. A `heap` has a size in bytes, and a location. The location can be local to the graphics device, or not local. In most systems, there will be two `heaps`: one on the Vulkan device, and one on the system used by the CPU. When you want to use some memory from a `heap`, you allocate it as a certain type of memory. Each type has different properties; a summary is below.
+
+*VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT*: If this is set, memory allocated with this type is the most efficient for device access. This bit _will only be set if the memory is also visible to the `host`_.
```
termhn: This bit only says that the memory is local to the *device* (or whatever memory is closest to it), and may or may not also be visible to the host.

docs/buffers.md
```diff
@@ -0,0 +1,69 @@
...
+
+1. A server with multiple graphics cards meant for high parallelizable computation (machine learning, neural net training, etc)
+2. CPUs with the GPUs embedded on the chip, also known as SOCs (System on a Chip)
+3. Devices with Unified Memory Access (UMA) where memory is non-local to _both_ the host and the device (weird, huh?)
+4. A GPU with no memory
+5. A desktop system with 2 or more graphics cards
+
+### Allocation and Deallocation
+
+When working with Vulkan, the programmer (that's you!), is responsible for requesting and freeing memory. This means it is possible to leak memory, even when using a safe language such as Rust.
```
termhn: Well, technically, only through usage of unsafe should this be able to happen. That's why most of the functions that are marked as 'unsafe' related to memory are marked as such

docs/buffers.md
```diff
@@ -0,0 +1,69 @@
...
+            usage: gfx_hal::buffer::Usage::UNIFORM
+                | gfx_hal::buffer::Usage::INDIRECT
+                | gfx_hal::buffer::Usage::VERTEX,
+        },
+        Dynamic,
+    )
+    .unwrap();
+```
+
+This creates a buffer with a size large enough to hold the data for multiple frames in our game, can store indirect draw commands, and can store vertex data. It also has the Rendy type of `Dynamic`, which means it can be used to send data back and forth between the CPU and GPU (bidirectional) rather than just one direction (unidirectional).
```
termhn: Dynamic just means that we intend to change the data frequently (every frame or thereabouts) and it basically just guarantees that the memory has `VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT`, i.e. that it is able to be mapped and written to without the use of a staging buffer/buffer copy operation being submitted to a queue.

docs/buffers.md
fhaynes: Yes, but still worth calling out, I think.

docs/buffers.md
fhaynes: It looks like it also requires that it not have the lazily allocated bit, and opts for GPU memory if available? The rendy types are trying to abstract several concepts, which is fine, but I think they need to be described at the same abstraction level. Even better would be use-cases for each type.

docs/buffers.md
fhaynes: It will only be set if the `VK_MEMORY_HEAP_DEVICE_LOCAL_BIT` is set. I apparently substituted HOST for HEAP when I read it. Fixed, thanks.

docs/buffers.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Memory Visibility and Coherency
+
+Can the CPU see the contents of the GPU's memory? Can the GPU access the system RAM? Good questions! The answer is, it depends. A graphics device will have capabilities, and it is up to you, intrepid programmer, to figure out what those are and code accordingly.
+
+A piece of memory is referred to as a `heap`. A `heap` has a size in bytes, and a location. The location can be local to the graphics device, or not local. In most systems, there will be two `heaps`: one on the Vulkan device, and one on the system used by the CPU. When you want to use some memory from a `heap`, you allocate it as a certain type of memory. Each type has different properties; a summary is below.
+
+*VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT*: If this is set, memory allocated with this type is the most efficient for device access. This bit _will only be set if the heap has the `VK_MEMORY_HEAP_DEVICE_LOCAL_BIT` set as well_.
```
omni-viral: will be set if and only if

docs/buffers.md
```diff
@@ -0,0 +1,92 @@
...
+
+*VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT*: If this is set, the memory is visible to the `host`; the system with the CPU. 
+
+*VK_MEMORY_PROPERTY_HOST_COHERENT_BIT*: If this is set, host writes or device writes become visible to each other without explicit flush commands
```
omni-viral: This wording may confuse user that it become visible automatically without any synchronization operations.
But only flushing and invalidating are not required in case of coherent memory type.

docs/buffers.md
```diff
@@ -0,0 +1,92 @@
...
+
+*VK_MEMORY_PROPERTY_HOST_CACHED_BIT*: If this is set, the memory is cached on the `host`. _This memory may not be host coherent, but is usually faster than uncached memory_. This means that cached memory may not reflect the latest writes to it.
```
omni-viral: User should avoid reading from non-cached memory types.
Writes does not use cache anyway. So it is not always faster.

docs/buffers.md
```diff
@@ -0,0 +1,92 @@
...
+
+*VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT*: If this is set, memory allocated with this type is the most efficient for device access. This bit _will only be set if the heap has the `VK_MEMORY_HEAP_DEVICE_LOCAL_BIT` set as well_.
+
+*VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT*: If this is set, the memory is visible to the `host`; the system with the CPU. 
+
+*VK_MEMORY_PROPERTY_HOST_COHERENT_BIT*: If this is set, host writes or device writes become visible to each other without explicit flush commands
+
+*VK_MEMORY_PROPERTY_HOST_CACHED_BIT*: If this is set, the memory is cached on the `host`. _This memory may not be host coherent, but is usually faster than uncached memory_. This means that cached memory may not reflect the latest writes to it. 
+
+*VK_MEMORY_PROPERTY_LAZILY_ALLOCATED_BIT*: If this is set, the memory is visible to the device only. Despite its name, whatever memory is behind this allocation may or may not be lazily allocated.
```
omni-viral: Add that memory can be used only for images with `VK_IMAGE_USAGE_TRANSIENT_ATTACHMENT_BIT` usage flag.

docs/buffers.md
omni-viral: Leaks are safe. You can always just forget buffer.

docs/buffers.md
omni-viral: No lazy allocated bit is required because it can't be host visible then.
It does indeed tries to use host-visible device-local memory available on integrated GPUs and some AMD GPUs (and maybe somewhere else) to minimize device access latency.

docs/buffers.md
```diff
@@ -0,0 +1,92 @@
...
+Below is a summary of the memory types offered by Rendy. These contain a combination of flags most useful for their purpose; in some cases, they are differentiated only by usage. That is, the underlying Vulkan flags may be the same, but the Rendy type differs in order to be specific about usage.
+
+The code for these can be found in `rendy/memory/usage.rs`.
+
+1. Dynamic
+    * Host visible
+    * _Prefers_ memory with fast GPU access
+    * Used for data that will be frequently updated from the system side, such as uniform buffers, indirect calls, or instance    rate vertex buffers
+2. Upload
+    * Host visible
```
omni-viral: Prefers not device local, not cached

docs/buffers.md
```diff
@@ -0,0 +1,92 @@
...
+    * CPU to GPU data flow with mapping
+    * Used for staging data before copying to `Data` memory
+3. Download
+    * Host visible
+    * _Prefers_ memory with fast GPU access
```
omni-viral: Prefers **not** device local, cached.

docs/buffers.md
```diff
@@ -0,0 +1,92 @@
...
+    * Used for copying data from `Data` memory to the host
+4. Data
+    * Device local
```
omni-viral: and not lazily allocated.

omni-viral:
Great job!
Couple of inaccuracies though :)

docs/buffers.md
fhaynes: Hrmm, what part do you think might cause confusion? I think it says what you said.

docs/buffers.md
fhaynes: These are meant to just describe the settings, not make suggestions about usage or not. I'll add that only reads are faster.

docs/buffers.md
fhaynes: But only when used for images, right?

docs/buffers.md
omni-viral: "A happens without B" can imply that B was the only requirement otherwise

docs/buffers.md
omni-viral: I guess image usage flag can be used only for images 😉

fhaynes: Hmm, the Vulkan documentation did not say that lazily allocated memory must always and only have the `VK_IMAGE_USAGE_TRANSIENT_ATTACHMENT_BIT` set. Can you give me a link to the Vulkan API sections so I can include it?

docs/buffers.md
fhaynes: I'm still uncertain how it is unclear. I think it is pretty clear that explicit flush commands are only unnecessary if that bit is set. How would you reword it to be clearer?

docs/buffers.md
omni-viral: Memory doesn't have `VK_IMAGE_USAGE_TRANSIENT_ATTACHMENT_BIT`. Image does

omni-viral: https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#memory-device-lazy_allocation

fhaynes: Yes, but is:
```
A memory type with this flag set is only allowed to be bound to a VkImage whose usage flags include VK_IMAGE_USAGE_TRANSIENT_ATTACHMENT_BIT.
```

also mean that this memory type can be bound to a VkImage and *only* a VkImage, or that when it is bound to a VkImage, that VkImage must have the transient attachment bit set? And if it is bound to something that is *not* a VkImage, it wouldn't apply?

fhaynes: Talked to @omni-viral, it can only be used with a VkImage.

PR #96, opened 2019-04-09
Share transient resources.

omni-viral:
Do not permit mutable access.


omni-viral:
bors retry

Frizi:
bors retry

PR #100, opened 2019-04-12
Improve `load_from_image` and mip map configuration

termhn:


texture/src/format/image.rs
```diff
@@ -239,59 +241,87 @@ macro_rules! dyn_format {
...
 
+    let (kind, layout) = config.kind.layout_and_kind(w, h);
+
+
     Ok(TextureBuilder::new()
         .with_raw_data(vec, format)
         .with_swizzle(swizzle)
-        .with_data_width(layout.line_stride)
-        .with_data_height(layout.layer_stride)
+        .with_data_width(w)
```
Frizi: I think it should still use layout (because of 1D), but it's calculation should be different. Also i would rename the strides back to `data_width` and `data_height`, as those old names were based on bad assumption :P

termhn:
ah, need to update examples...

termhn:
I think this is ready for review as it now works properly and is being used in `rendy-pbr` ibl branch. I updated example and builds passing :D

termhn:
bors r=Frizi

termhn:
bors r=Frizi

PR #105, opened 2019-04-18
Mipmap support, Blitter and Uploader refactor

Frizi:
Mipmaps are now autogenerated for all textures ~~unless explicitly disabled~~ based on TextureBuilder config, disabled by default.

Added batching to Uploader barriers. All transfer barriers and transitions are batched together inside the flush.

The lack of barrier batching is not ideal for blitter, but i think it's optimal in case of mipmaps anyway, due to all levels depending on previously generated ones. We need something more involved than that to support more general case optimally, but that's a problem for another day.

### DIFF
--- a/factory/src/barriers.rs
+++ b/factory/src/barriers.rs
@@ -3,8 +3,8 @@ use {
         command::Encoder,
         resource::{Handle, Image},
     },
-    gfx_hal::{self, buffer, image, pso, Backend},
-    std::{iter::once, ops::Range},
+    gfx_hal::{self, buffer, image, memory::Barrier, pso, Backend},
+    std::ops::Range,
 };
 
 /// A variant of `gfx_hal::image::Barrier` that uses Handle<Image<B>>
@@ -21,8 +21,8 @@ struct ImageBarrier<B: Backend> {
 }
 
 impl<B: Backend> ImageBarrier<B> {
-    fn raw(&self) -> gfx_hal::memory::Barrier<'_, B> {
-        gfx_hal::memory::Barrier::Image {
+    fn raw(&self) -> Barrier<'_, B> {
+        Barrier::Image {
             states: self.states.clone(),
             target: self.target.raw(),
             families: None,
@@ -123,16 +123,16 @@ impl<B: Backend> Barriers<B> {
     }
 
     pub fn encode_before<C, L>(&mut self, encoder: &mut Encoder<'_, B, C, L>) {
-        if self.before_stages != pso::PipelineStage::empty() {
+        if !self.before_stages.is_empty() {
             let transitions = self.before_image_transitions.iter().map(|b| b.raw());
-            let all_images = once(gfx_hal::memory::Barrier::AllImages(
+            let all_images = Some(Barrier::AllImages(
                 self.before_image_access..self.target_image_access,
             ))
-            .filter(|_| self.before_image_access != image::Access::empty());
-            let all_buffers = once(gfx_hal::memory::Barrier::AllBuffers(
+            .filter(|_| !self.before_image_access.is_empty());
+            let all_buffers = Some(Barrier::AllBuffers(
                 self.before_buffer_access..self.target_buffer_access,
             ))
-            .filter(|_| self.before_buffer_access != buffer::Access::empty());
+            .filter(|_| !self.before_buffer_access.is_empty());
 
             encoder.pipeline_barrier(
                 self.before_stages..self.target_stages,
@@ -150,16 +150,16 @@ impl<B: Backend> Barriers<B> {
     }
 
     pub fn encode_after<C, L>(&mut self, encoder: &mut Encoder<'_, B, C, L>) {
-        if self.target_stages != pso::PipelineStage::empty() {
+        if !self.target_stages.is_empty() {
             let transitions = self.after_image_transitions.iter().map(|b| b.raw());
-            let all_images = once(gfx_hal::memory::Barrier::AllImages(
+            let all_images = Some(Barrier::AllImages(
                 self.target_image_access..self.after_image_access,
             ))
-            .filter(|_| self.after_image_access != image::Access::empty());
-            let all_buffers = once(gfx_hal::memory::Barrier::AllBuffers(
+            .filter(|_| !self.after_image_access.is_empty());
+            let all_buffers = Some(Barrier::AllBuffers(
                 self.target_buffer_access..self.after_buffer_access,
             ))
-            .filter(|_| self.after_buffer_access != buffer::Access::empty());
+            .filter(|_| !self.after_buffer_access.is_empty());
 
             encoder.pipeline_barrier(
                 self.target_stages..self.after_stages,
--- a/factory/src/barriers.rs
+++ b/factory/src/barriers.rs
@@ -1,0 +1,241 @@
+use {
+    crate::{
+        command::Encoder,
+        resource::{Handle, Image},
+    },
+    gfx_hal::{self, pso::PipelineStage, Backend},
+    std::{
+        iter::once,
+        ops::{BitOrAssign, Range},
+    },
+};
+
+#[derive(Debug)]
+struct ImageBarrier<B: Backend> {
+    /// The access flags controlling the image.
+    pub states: Range<gfx_hal::image::State>,
+    /// The image the barrier controls.
+    pub target: Handle<Image<B>>,
+    /// The source and destination Queue family IDs, for a [queue family ownership transfer](https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#synchronization-queue-transfers)
+    /// Can be `None` to indicate no ownership transfer.
+    pub families: Option<Range<gfx_hal::queue::QueueFamilyId>>,
+    /// A `SubresourceRange` that defines which section of an image the barrier applies to.
+    pub range: gfx_hal::image::SubresourceRange,
+}
+
+impl<B: Backend> ImageBarrier<B> {
+    fn raw(&self) -> gfx_hal::memory::Barrier<'_, B> {
+        gfx_hal::memory::Barrier::Image {
+            states: self.states.clone(),
+            target: self.target.raw(),
+            families: self.families.clone(),
+            range: self.range.clone(),
+        }
+    }
+}
+
+#[derive(Debug)]
+pub(crate) struct ImgBarrierCollector<B: Backend> {
+    target_stage: PipelineStage,
+    target_access: gfx_hal::image::Access,
+    target_layout: gfx_hal::image::Layout,
+    before_barriers: (Option<PipelineStage>, Vec<ImageBarrier<B>>),
+    after_barriers: (Option<PipelineStage>, Vec<ImageBarrier<B>>),
+    before_global: Option<(PipelineStage, gfx_hal::image::Access)>,
+    after_global: Option<(PipelineStage, gfx_hal::image::Access)>,
+}
+
+impl<B: Backend> ImgBarrierCollector<B> {
+    pub fn new(
+        target_stage: PipelineStage,
+        target_access: gfx_hal::image::Access,
+        target_layout: gfx_hal::image::Layout,
+    ) -> Self {
+        Self {
+            target_stage,
+            target_access,
+            target_layout,
+            before_barriers: (None, Vec::new()),
+            after_barriers: (None, Vec::new()),
+            before_global: None,
+            after_global: None,
+        }
+    }
+
+    fn add_last_access(&mut self, stage: PipelineStage, access: gfx_hal::image::Access) {
+        merge_opt(&mut self.before_global, (stage, access));
+    }
+
+    fn add_next_access(&mut self, stage: PipelineStage, access: gfx_hal::image::Access) {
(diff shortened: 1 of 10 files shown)


factory/src/upload.rs
```diff
@@ -174,27 +181,19 @@ where
...
-                    target: buffer.raw(),
-                    families: None,
-                    range: range.clone(),
-                }),
-            );
         }
 
+        family_uploads
+            .barriers
+            .add_buffer(last.map(|l| (l.stage, l.access)), (next.stage, next.access));
```
omni-viral: I just realized we can specify `Access::empty()` always as `last_access` for buffers
and for images if no layout transition required.

factory/src/upload.rs
```diff
@@ -254,63 +241,54 @@ where
...
+                    gfx_hal::image::Access::empty(),
+                    last_layout,
+                )
             }
-        }
+        };
 
+        let target_layout = match (last_layout, next.layout) {
+            (_, gfx_hal::image::Layout::General) => gfx_hal::image::Layout::General,
+            (gfx_hal::image::Layout::General, _) => gfx_hal::image::Layout::General,
```
omni-viral: If `next.layout` or `last_layout` is `TransferDstOptimal` then target layout could be `TransferDstOptimal` as well, even if the other one is `General`.

Besides, we want to eliminate transition before update rather than after.
Which leads us to this:
```
let target_layout = match (last_layout, next_layout) {
  (TransferDstOptimal, _) => TransferDstOptimal,
  (General, TransferDstOptimal) => TransferDstOptimal, // I'm uncertain here. If before transition would be elimintated we will possibly relax access mask and thus could remove unncecessary cache flush.
  (_, General) => General,
  (General, _) => General,
  _ => TransferDstOptimal,
};
```

factory/src/upload.rs
```diff
@@ -385,21 +347,26 @@ where
 #[derive(Debug)]
 pub(crate) struct FamilyUploads<B: gfx_hal::Backend> {
     pool: CommandPool<B, Transfer, IndividualReset>,
+    barrier_buffers: Vec<CommandBuffer<B, Transfer, InitialState, PrimaryLevel, IndividualReset>>,
```
omni-viral: We don't need to store them separately.

factory/src/upload.rs
```diff
@@ -446,17 +421,22 @@ where
         match &mut self.next[queue] {
             Some(next) => Ok(next),
             slot @ None => {
-                let buffer = self.command_buffers.pop().map(Ok).unwrap_or_else(
-                    || -> Result<_, failure::Error> { Ok(pool.allocate_buffers(1).pop().unwrap()) },
-                )?;
+                let command_buffer = self
+                    .command_buffers
+                    .pop()
+                    .unwrap_or_else(|| pool.allocate_buffers(1).pop().unwrap());
```
omni-viral: Just allocate two 😉

texture/src/texture.rs
```diff
@@ -47,6 +47,13 @@ where
     }
 }
 
+#[derive(Clone, Copy, Debug)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub enum MipLevel {
+    Auto,
+    Level(u8),
```
omni-viral: `NonZero`? :)

factory/src/blitter.rs
```diff
@@ -0,0 +1,423 @@
...
+        let mut src_next = next_iter.next().unwrap();
+        assert_eq!(src_last.queue, src_next.queue);
+
+        for (level, (dst_last, dst_next)) in (1..image.levels())
+            .into_iter()
+            .zip(last_iter.zip(next_iter))
+        {
+            assert_eq!(dst_last.queue, dst_next.queue);
+
+            let begin = level == 0;
```
omni-viral: level starts at `1` in this loop.

factory/src/blitter.rs
```diff
@@ -0,0 +1,423 @@
...
+                src: BlitImageState {
+                    subresource: gfx_hal::image::SubresourceLayers {
+                        aspects,
+                        level: level - 1,
+                        layers: 0..image.layers(),
+                    },
+                    bounds: gfx_hal::image::Offset::ZERO
+                        .into_bounds(&image.kind().level_extent(level - 1)),
+                    last_stage: if begin { src_last.stage } else { transfer },
+                    last_access: if begin { src_last.access } else { read },
```
omni-viral: `last_access` is `TRANSFER_WRITE` for levels > 0 when they are acting like a source.

factory/src/blitter.rs
```diff
@@ -0,0 +1,423 @@
...
+                    subresource: gfx_hal::image::SubresourceLayers {
+                        aspects,
+                        level,
+                        layers: 0..image.layers(),
+                    },
+                    bounds: gfx_hal::image::Offset::ZERO
+                        .into_bounds(&image.kind().level_extent(level)),
+                    last_stage: dst_last.stage,
+                    last_access: dst_last.access,
+                    last_layout: dst_last.layout,
```
omni-viral: Should be `Undefined` as we overwrite whole image level.

factory/src/blitter.rs
```diff
@@ -0,0 +1,423 @@
...
+            next,
+            read_barriers,
+            write_barriers,
+            ..
+        } = family_ops.deref_mut();
+
+        let next_ops = next[queue_id.index].as_mut().unwrap();
+        let mut encoder = next_ops.command_buffer.encoder();
+
+        let regions = regions
```
omni-viral: If multiple regions of the same level and layer get blitted we shoudn't sync that `SubresourceRange` more than once.

factory/src/barriers.rs
```diff
@@ -0,0 +1,188 @@
...
+            log::trace!(
+                "Transition last: {:?}",
+                (last_access, last_layout)..(self.target_image_access, target_layout)
+            );
+            self.before_image_transitions.push(ImageBarrier {
+                states: (last_access, last_layout)..(self.target_image_access, target_layout),
+                target: image.clone(),
+                range: image_range.clone(),
+            });
+        } else if !last_access.contains(self.target_image_access) {
```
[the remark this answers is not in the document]
Frizi: If there i no format transition and the access already matches, we don't need a barrier.

factory/src/barriers.rs
```diff
@@ -0,0 +1,188 @@
...
+        }
+    }
+
+    pub fn add_buffer(
+        &mut self,
+        last: Option<(pso::PipelineStage, gfx_hal::buffer::Access)>,
+        next: (pso::PipelineStage, gfx_hal::buffer::Access),
+    ) {
+        if let Some(last) = last {
+            if !last.1.contains(self.target_buffer_access) {
```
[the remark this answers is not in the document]
Frizi: Same as above. No need for a barrier as the access is already right.

factory/src/barriers.rs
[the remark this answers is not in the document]
Frizi: This is of course wrong. Fixed.

factory/src/blitter.rs
Frizi: This i will have to leave for later.

Frizi:
> If more specialized module would lead to significantly better performance and cleaner design
I agree that we might want that, but i like working solution infinitely more than no solution. We can make the optimization later.

factory/src/blitter.rs
omni-viral: Can you assert that this doesn't happen then? Because it won't just oversynchronize, it will to invalid image transitions.

Frizi: I have added a safety note for now. We will probably lift that issue later, but having a runtime debug check is already half-way to full implementation :P So i'm leaving that for later.

texture/src/format/image.rs
```diff
@@ -80,18 +43,13 @@ pub enum TextureKind {
     D2Array {
         samples: u8,
         layers: u16,
-        layout: LayerLayout,
     },
     D3 {
```
omni-viral: I guess this means that 2d image should be interpreted as several "layers" (similar to `2DArray`) and image `height` should be divided by `depth`.
Please, add short documentation for that to not require reader to figure it from code each time :)

Also, as the enum is dedicated to be used for loading textures from 2D images, it should be called accordingly with semantics. Because in general `TextureKind::3D { depth: 42 }` makes little sense.

Frizi: What do you mean that it makes little sense? It's perfectly valid to load such thing.

texture/src/format/image.rs
```diff
@@ -80,18 +43,13 @@ pub enum TextureKind {
     D2Array {
         samples: u8,
```
omni-viral: Can loaded textures be multisampled?

texture/src/format/image.rs
```diff
@@ -101,54 +59,53 @@ impl TextureKind {
...
                 D2(
-                    layout.layer_width(width, 6),
-                    layout.layer_height(height, 6),
+                    width,
+                    height / 6,
                     6,
                     1,
                 ),
-                layout.data_layout(width, height, 6),
+                DataLayout { width, height: height / 6 }
```
omni-viral: It seems that now `DataLayout`'s `width` and `height` are equal to `image::Kind`'s `width` and `height`.

texture/src/format/image.rs
omni-viral: That's exactly what I mean. It makes sense for loading 3D image from 2D image. i.e. when there is something that tells us about `width` and `height`. But name `TextureKind` does not imply that there is 2D image to fetch missing info.

texture/src/format/image.rs
Frizi: This doesn't seem right. I think there might be some arcane ways to do it, but I don't think this needs to be exposed through texture builder.

texture/src/format/image.rs
Frizi: Ok. I've added documentation and also simplified the type.

texture/src/format/image.rs
Frizi: Yup, removed the redundancy.

factory/src/barriers.rs
```diff
@@ -0,0 +1,178 @@
...
+        self.before_image_transitions.clear();
+    }
+
+    pub fn encode_after<C, L>(&mut self, encoder: &mut Encoder<'_, B, C, L>) {
+        if self.target_stages != pso::PipelineStage::empty() {
+            let transitions = self.after_image_transitions.iter().map(|b| b.raw());
+            let all_images = once(gfx_hal::memory::Barrier::AllImages(
+                self.target_image_access..self.after_image_access,
+            ))
+            .filter(|_| self.after_image_access != image::Access::empty());
```
omni-viral: Can you make this into `Option<Barrier>`?
Like this for example.
```
if self.after_image_access.is_empty() { None } else { Some(Barrier::AllImages(self.after_image_access)) }
```

factory/src/blitter.rs
```diff
@@ -0,0 +1,426 @@
...
+                    gfx_hal::pso::PipelineStage::TRANSFER,
+                    gfx_hal::buffer::Access::TRANSFER_WRITE,
+                    gfx_hal::image::Access::TRANSFER_WRITE,
+                ),
+            }));
+        }
+
+        Ok(Blitter { family_ops })
+    }
+    /// `dst` should be `None` when blitting from the same image.
```
omni-viral: What's `dst`?

factory/src/blitter.rs
```diff
@@ -0,0 +1,426 @@
...
+
+            log::trace!("Blit: {:#?}", blit);
+            self.blit_image(device, src_last.queue, &image, &image, filter, Some(blit))?;
+            src_last = dst_last;
+            src_next = dst_next;
+        }
+        Ok(())
+    }
+
+    /// `dst` should be `None` when blitting from the same image.
```
omni-viral: Again. What's `dst`?

factory/src/blitter.rs
```diff
@@ -0,0 +1,426 @@
...
+                    dst_bounds: reg.dst.bounds,
+                }
+            })
+            .collect::<SmallVec<[_; 1]>>();
+
+        // TODO: synchronize whatever possible on flush.
+        // Currently all barriers are inlined due to dependencies between blits.
+
+        read_barriers.encode_before(&mut encoder);
+        write_barriers.encode_before(&mut encoder);
```
omni-viral: Can't we record only read barriers before and only write barriers after blit op?

factory/src/blitter.rs
Frizi: Bad docs, fixed

factory/src/barriers.rs
Frizi: What's the difference? I can literally change `once` call to `Some` variant and it will be exactly the same. I'd prefer to stick with iterators here, because everything else here is an iterator and all of those are finally chained together.

omni-viral: `Option` is `IntoIterator`.
I just suggest to change type of the iterator here.
From `Filter<Once<T>>` to `Option<T>`

torkleyy:
Seems like a spurious failure.

bors r=omni-viral

PR #109, opened 2019-04-24
Adds a document covering the basics of the render graph

fhaynes:


docs/graph.md
```diff
@@ -0,0 +1,55 @@
+# Summary
+
+This document gives an overview of the Framegraph system in Rendy. 
+
+## What is a Framegraph
+
+Let's start at the end for this explanation. What you see on your monitor is the end result of a complicated series of transformations to data as it makes its way from your running application to your monitor.
+
+This series of transformations is often referred to as a _rendering pipeline_. These can be very simple, or they can be very complex. When they are very complex, it can be useful to add a layer of abstraction to make them easier to conceptualize.
```
Frizi: The `rendering pipeline` is also often used to refer to the gpu hardware pipeline (all shader stages and things inbetween). It might sounds confusing here for people familiar with the term

docs/graph.md
```diff
@@ -0,0 +1,55 @@
...
+
+The render graph just wraps them up neatly into a
+
+## Render Group
+
+A render group is a collection of pipelines (in the Vulkan sense of the word) that are processed in a `Subpass`.
+
+## Subpass
+
+A `Subpass` is a child of a `Pass`. Currently, a `Pass` can have only one `Subpass`. While useful, they are not required. They are an opportunity for performance optimization when the output of one render group is used directly as the input to another one; not sampled, but literally passing the fragment.
```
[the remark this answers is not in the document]
omni-viral: You can add that `Subpass` concept in API exists because once "1 subpass per pass" restriction is lifted there won't be any API breakage.

docs/graph.md
```diff
@@ -0,0 +1,55 @@
...
+
+This series of transformations is often referred to as a _rendering pipeline_. These can be very simple, or they can be very complex. When they are very complex, it can be useful to add a layer of abstraction to make them easier to conceptualize.
+
+One way of doing this is with a _framegraph_ (an alternate name sometimes used is _render graph_). Consider a simple rendering pipeline that looks like this:
+
+```
+PBR Shader -> Dither -> Lens Flare
+```
+
+This could be applied to one model in your frame. Expand this out to much more complex pipelines, across complex scenes with many models, and you can see how it can get complicated.
```
termhn: Adding an additional example here of a situation in which a graph splits more would be useful I think. The linear situation above seems quite simple and might not illustrate where the graph helps. One example of this could be something like the following...

```
Previous frame -> Compute average brightness-\
                                              \
                                               v
Depth pre-pass -> PBR Shader -> Dither -> Tonemap -> Lens flare -> Output
                    ^
                   /
Shadow map pass --/
```

fhaynes: Ah, thanks!

docs/graph.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Synchronization
+
+Aside from providing a higher-level view of your graphics pipeline, a render graph can also handle _synchronization_. Imagine you have 3 shaders, A, B, and C. C requires both A and B to have processed a pixel before it can process it.
+
+A graph can abstract the synchronization primitives needed to ensure C doesn't get the data before both A and B are done.
+
+### Ordering
+
+Similar to synchronization, a render graph can provide ordering. If you have shaders A -> B -> C, you may not want them to ever run in the order B -> C -> A.
```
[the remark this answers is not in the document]
fhaynes: Well, there's two cases...one where you want everything up to a point to happen in a specific order:

A -> B -> C -> Sync point

This is when B relies on A, and C relies on B. But you can also have C that relies on A and B, but A and B are disjunct, so it doesn't matter what order they arrive at C in, C will wait until both are there.

docs/graph.md
fhaynes: good point, thanks

docs/graph.md
```diff
@@ -0,0 +1,75 @@
...
+
+The render graph just wraps them up neatly into a
+
+## Render Group
+
+A render group is a collection of pipelines (in the Vulkan sense of the word) that are processed in a `Subpass`.
+
+## Subpass
+
+A `Subpass` is a child of a `Pass`. Currently, a `Pass` can have only one `Subpass`. While useful, they are not required. They are an opportunity for performance optimization when the output of one render group is used directly as the input to another one; not sampled, but literally passing the fragment.
```
omni-viral: One `Subpass` is always required.

docs/graph.md
```diff
@@ -0,0 +1,75 @@
...
+
+An additional benefit to having and using `Subpasses` now is that the API will not require breaking changes to support more than one `Subpass` per `Pass`. 
+
+## Pass
```
omni-viral: We call it `RenderPass` usually

docs/graph.md
```diff
@@ -0,0 +1,75 @@
...
+
+A `Pass` currently can have only one `Subpass`, and a `Pass` will usually belong to a `Node`.
```
omni-viral: `RenderPass` is a `Node`.

docs/graph.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Node
+
+A `Node` contains 0 or more things that provide a set of self-contained (i.e. internally synchronized) submissions to a queue each frame. This is _usually_ a `Pass`, but does not have to be.
```
omni-viral: While technically this is correct, but intention is that `Node` to be exactly one self-contained thing.

docs/graph.md
fhaynes: Fixed

PR #113, opened 2019-04-26
Change all github links to amethyst/rendy

icefoxen:
I THINK this is right.  I keep getting annoyed that all the crates.io repo links take me to what appears to be an old repo.


omni-viral:
We transferred repo just few days ago and didn't fixed links yet, so thanks!

Issue #111, opened 2019-04-25
Re-export `failure`


icefoxen:
The `SimpleGraphicsPipeline` trait has a method that returns `failure::Error`: https://docs.rs/rendy/0.1.1/rendy/graph/render/trait.SimpleGraphicsPipeline.html#tymethod.build

This seems unnecessary since it looks like it could be easy to replace with an associated `Error` type, but if we want to use `failure` we should at least re-export it similar to how `gfx_hal` is re-exported.  That way if users have a version clash somehow, they at least have a convenient reference to the version of `failure` we are using.  This honestly isn't likely, given how stable `failure` is, but I find it to be a useful policy.

omni-viral:
Associated error type would do, but different error types from different pipelines still have to be merged.
But reexporting looks like a good idea.

malobre:
Since #211 was merged maybe this should be closed?

PR #114, opened 2019-04-26
Update piles of minor dependencies.

icefoxen:
...mainly out of boredom.

This does have ONE significant change: It removes the `parallel` feature from `hibitset`.  This feature uses `rayon`, which feels relatively heavyweight to me.  More to the point, it pulls in a slightly different version of `crossbeam`, which results in duplicate deps of different versions, which makes my OCD tweak every time I build from scratch.

If we want that feature enabled, I'll happily fix that, or make a feature flag that lets one choose.  Up to you guys, I'm not sure what the implications are.

### DIFF
--- a/resource/src/escape.rs
+++ b/resource/src/escape.rs
@@ -70,7 +70,12 @@ impl<T> Drop for Escape<T> {
     fn drop(&mut self) {
         unsafe {
             // Read value from `ManuallyDrop` wrapper and send it over the channel.
-            self.sender.send(read(&mut *self.value));
+            match self.sender.send(read(&mut *self.value)) {
+                Ok(_) => {}
+                Err(_) => {
+                    log::error!("`Escape` was dropped after a `Terminal`?");
+                }
+            }
         }
     }
 }
@@ -114,7 +119,7 @@ impl<T> Terminal<T> {
         repeat(()).scan(&mut self.receiver, move |receiver, ()| {
             // trace!("Drain escape");
             if !receiver.is_empty() {
-                receiver.recv()
+                receiver.recv().ok()
             } else {
                 None
             }
@@ -127,8 +132,8 @@ impl<T> Drop for Terminal<T> {
         unsafe {
             ManuallyDrop::drop(&mut self.sender);
             match self.receiver.try_recv() {
-                None => {}
-                Some(_) => {
+                Ok(_) => {}
+                Err(_) => {
                     log::error!("Terminal must be dropped after all `Escape`s");
                 }
             }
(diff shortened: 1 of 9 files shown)


icefoxen:
Oh, and the one thing I didn't bump was `shaderc`, since that seems a relatively Big Deal and I'm a little scared to touch it.  Doing some investigation though, the 0.3 -> 0.5 transition appears to mainly involve refactoring out the C parts into a `-sys` crate, and the result builds cleanly on my (linux) system.  Can't test the actual graphics just at the moment.

...and the `winit` version bump may also be a bad idea since `gfx_hal` 0.1 still uses `winit` 0.18 it seems.  Not sure why that built in the first place.

omni-viral:
@icefoxen thanks.
Although winit version bump seems to break build.

omni-viral:
For winit and other remote types we use to communicate with gfx-hal have to be the same version.
I think gfx-hal should update its winit dependency (and probably everything where update is available prior 0.2)

icefoxen:
Yeah, winit has been downgraded again and now all tests appear to pass on my machine.  It'd be nice if there were a single place we can describe these version dependencies...  maybe gfx-hal can re-export `winit` a la #111 ?  Or am I just seeing every problem as a nail?

omni-viral:
Reexporting everything is not feasible.
Also `rendy` does not produce `Window`, but receives it, so it can be created by another crate that does not depend on `rendy` at all. So we only can require version match.
I wish pub dependencies will be a thing soon.

resource/src/escape.rs
```diff
@@ -127,8 +132,8 @@ impl<T> Drop for Terminal<T> {
         unsafe {
             ManuallyDrop::drop(&mut self.sender);
             match self.receiver.try_recv() {
-                None => {}
-                Some(_) => {
+                Ok(_) => {}
```
omni-viral: Should be vice versa.

icefoxen: Good catch, thank you.

icefoxen: Wait, should it?  Is it checking to make sure it *can't* receive from anything and printing a warning if it can?  Just want to double check.

omni-viral: It checks that all `Senders` are dropped.
So it must warn if function returns anything else than `Err(Disconnected)`.

icefoxen: Makes sense.  Fixed.

jaynus:
Just to note, shaderc was bumped in #108 so its a non-issue

PR #120, opened 2019-04-30
Add window centering.

mralve:
Ahh after a long journey... finally this is the right repo, lol.

Added window centering functions such as center(), set_center_monitor(), set_centered() in wsi::WindowExt

this PR can be reverted if my PR (which provides the same functionality) on winit gets accepted.

### DIFF
--- a/rendy/examples/meshes/main.rs
+++ b/rendy/examples/meshes/main.rs
@@ -21,6 +21,7 @@ use {
         mesh::{AsVertex, Mesh, PosColorNorm, Transform},
         resource::{Buffer, BufferInfo, DescriptorSet, DescriptorSetLayout, Escape, Handle},
         shader::{Shader, ShaderKind, SourceLanguage, SpirvShader, StaticShaderInfo},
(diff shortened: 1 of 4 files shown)


termhn:
Seems to me like this doesn't really belong in `rendy` but rather in `winit` itself since this functionality is rather orthogonal to rendering itself? `wsi` is basically just meant to provide minimum necessary interaction with windowing library in order to get it to integrate with gfx-hal/vulkan

mralve:
@termhn No your right, I currently have a PR in Winit now, that is under review. so it will move there instead. wonder how it looks with winit in Rendy right now? winit got reverted to 0.18 some days ago, is there any major problems currently?

torkleyy:
I think that revert was because gfx hasn't updated yet.

Frizi:
We use winit 0.19 in `hal-master` branch, which gfx-hal master.

Frizi:
winit has been updated alongside with hal 0.2 :)

PR #121, opened 2019-05-01
Support separate vertex buffers and bind them based on format and name

Frizi:
This is tied to https://github.com/amethyst/amethyst/issues/1558 and https://github.com/amethyst/amethyst/issues/1557


mesh/src/mesh.rs
```diff
@@ -205,101 +212,129 @@ impl<'a> MeshBuilder<'a> {
...
-                                        .with_stage(gfx_hal::pso::PipelineStage::VERTEX_INPUT),
-                                )?;
-                            }
-                            buffer
-                        },
-                        format: format.clone(),
-                    })
+        let align = factory.physical().limits().non_coherent_atom_size;
+
+        let mut len = self
```
termhn: validate all vertices are the same len or put a comment explaining the behavior

rendy/examples/meshes/main.rs
```diff
@@ -112,23 +114,23 @@ struct Aux<B: gfx_hal::Backend> {
 const MAX_LIGHTS: usize = 32;
 const MAX_OBJECTS: usize = 10_000;
 const UNIFORM_SIZE: u64 = size_of::<UniformArgs>() as u64;
-const TRANSFORMS_SIZE: u64 = size_of::<Transform>() as u64 * MAX_OBJECTS as u64;
+const MODELS_SIZE: u64 = size_of::<Model>() as u64 * MAX_OBJECTS as u64;
```
termhn: I don't like this rename... I think Transform is more accurately describing what it is. Or even just Matrix could be better. Since it doesn't necessarily need to be a "Model" matrix, and I think that in this position Model is sort of a non-obvious/confusing name.

util/src/types/vertex.rs
```diff
@@ -397,114 +557,71 @@ pub struct PosNormTangTex {
...
 }
 
 /// Full vertex transformation attribute.
 /// Typically provided on per-instance basis.
 /// It takes 4 attribute locations.
 #[repr(transparent)]
 #[derive(Clone, Copy, Debug, PartialEq, PartialOrd)]
-pub struct Transform(pub [[f32; 4]; 4]);
-impl<T> From<T> for Transform
+pub struct Model(pub [[f32; 4]; 4]);
```
termhn: I guess this is the real place the rename happens

mesh/Cargo.toml
```diff
@@ -29,4 +29,8 @@ failure = "0.1"
 serde = { version = "1.0", optional = true, features = ["derive"] }
 wavefront_obj = { version = "6.0", optional = true }
 smallvec = { version = "0.6" }
-serde_bytes = { version = "0.11", optional = true }
+serde_bytes = { version = "0.10", optional = true }
```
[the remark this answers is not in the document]
termhn: nvm, this is from earlier review

rendy/examples/meshes/main.rs
Frizi: It does necessarily have to be "model", as it's associated "NAME" must be identical to one in the shader. Would you rather change the "model" shader attribute name to "transform"?

mesh/src/mesh.rs
Frizi: It doesn't differ from old behaviour, except using less smaller buffers on such odd cases.

Frizi: I have added a doc comment to the build function.

rendy/examples/meshes/main.rs
termhn: Ah, hm, fair enough. Alright, I'm fine with this then.

Frizi:
bors r=termhn,omni-viral

PR #125, opened 2019-05-12
New dynamic allocation strategy

omni-viral:


omni-viral:
@fhaynes Is this a question? I didn't add them.
Tests would require either mock gfx backend or making allocator gfx agnostic. Any of this is out of scope for this PR.

fhaynes:
> @fhaynes Is this a question? I didn't add them.
> Tests would require either mock gfx backend or making allocator gfx agnostic. Any of this is out of scope for this PR.

More that I want to make sure they get added soon. I know we have https://github.com/amethyst/rendy/issues/116, but I'll make a ticket for an agnostic gfx backend.

memory/src/allocator/dynamic.rs
```diff
@@ -198,105 +190,244 @@ where
...
+            // Allocate block for the chunk.
+            let (block, allocated) = self.alloc_from_entry(device, chunk_size, 1)?;
+            Ok((Chunk::from_block(block_size, block), allocated))
         } else {
-            // Allocate from larger chunks.
-            let (dynamic_block, allocated) = self.alloc_from_chunk(device, size)?;
-            Ok((Chunk::Dynamic(dynamic_block), allocated))
+            let total_blocks = self.sizes[&block_size].total_blocks;
+            let chunk_size =
+                (max_chunk_size.min(min_chunk_size.max(total_blocks * block_size)) / 2 + 1).next_power_of_two();
```
Frizi: shoudn't this be `- 1`?

factory/src/config.rs
```diff
@@ -169,24 +170,18 @@ unsafe impl HeapsConfigure for BasicHeapsConfigure {
                         .contains(gfx_hal::memory::Properties::CPU_VISIBLE)
                     {
                         Some(LinearConfig {
-                            linear_size: min(_256mb, properties.memory_heaps[mt.heap_index] / 8),
+                            linear_size: min(_128mb, properties.memory_heaps[mt.heap_index] / 16),
```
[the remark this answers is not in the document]
omni-viral: It's equally magic as `_128mb`. It basically mean `don't allocate linear chunks more than `1/16th` of the heap size.

memory/src/allocator/dynamic.rs
omni-viral: No. `(x / 2 + 1).next_power_of_two()` is `x.prev_power_of_two()`, except there is no `prev_power_of_two()` function 🙁

factory/src/config.rs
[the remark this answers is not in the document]
Frizi: I had a hard time guessing where that 16 comes from. The 128mb is kinda understandable. It seems it's just arbitrarily chosen too after all :P

memory/src/allocator/dynamic.rs
Frizi: `(64 / 2 + 1).next_power_of_two()` = 64. I'm not sure if that's what you expected. Still seems like the `+1` should go to match the behaviour of our usual `next_power_of_two()`s.

omni-viral: @Frizi Yes. This is exactly expected result.
I need largest power of two that is **non-greater** than value.

omni-viral: `(x - 1).next_power_of_two()` is actually wrong, because it results in `x-1` if `x-1` is power of two.
I thought that `next_power_of_two` result is always greater than argument. i.e. `2.next_power_of_two() == 3` but this is not the case.

omni-viral:
bors retry

PR #126, opened 2019-05-15
Expand docs for Mesh::len() a little.

### LINKED ISSUE #119: `Mesh::len()` is weird
It appears to be the number of vertices *that will be drawn* in a mesh.  So for a mesh with indices, this is the number of indices, and in a mesh without indices, this is the number of vertices.  I think.

For drawing this is generally the value we want, but it's kinda weird that it measure the length of two different things depending on context.  Does anyone think it might be nicer to have `verts_len()` and `indices_len()` be separate things or such?  Or should we just update the docs to make it more clear what it's actually measuring?

icefoxen:
Resolves #119

### DIFF
--- a/mesh/src/mesh.rs
+++ b/mesh/src/mesh.rs
@@ -358,7 +358,11 @@ where
         self.prim
     }
 
-    /// Get number of vertices in mesh.
+    /// Returns the number of vertices that will be drawn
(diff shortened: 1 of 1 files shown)


omni-viral:
LGTM bors r+

PR #130, opened 2019-05-17
Updated obj loader for seperate vertex format and multi objects.

valkum:
Not sure if this is a solution suitable for rendy but yesterday I came up with this updated obj loader inspired by the API of the gltf mesh loader.
Returns one MeshBuilder for each group. This way we can integrate mtl loading.
Currently each face has 3 vertices but in a future release we can remove duplicates and use the indexbuffer.

### DIFF
--- a/mesh/Cargo.toml
+++ b/mesh/Cargo.toml
@@ -30,3 +30,5 @@ serde = { version = "1.0", optional = true, features = ["derive"] }
 wavefront_obj = { version = "6.0", optional = true }
 smallvec = { version = "0.6" }
 serde_bytes = { version = "0.11", optional = true }
+
+log = "0.4.6"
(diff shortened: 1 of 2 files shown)


Frizi:
Any specific reason why you haven't used indices yet? Should be very straightforward, given that the api gives you indices and other buffers directly.

Frizi:
Anyway, just to make things faster, let's get it in asap

bors r+

valkum:
Tried that first but had some bugs. 

The problem is that the obj file uses a single vertex with multiple normals/texcoords. 
Thus it has multiple indicies to create the shapes. Because we only can use one index for a specific combination of (position, normal, texcoord) a simple usage of their indices is not possible.
We need to create each combination and then build our own indicies.

PR #133, opened 2019-05-18
Cleanup Cargo.toml files

### LINKED ISSUE #39: Make wsi winit agnostic
My use case requires the wayland layer shell protocol. Winit only works when you are building a standard xdg shell application.

The only reason winit is needed is to get the window size and hidpi scale factor, it should be possible to provide this information as a `size: (u32, u32)` when calling `into_target` and creating a new surface by providing a raw `gfx_hal::Backend::Surface`. The `Factory` should then either expose the `Instance` or take a closure that takes an `Instance` and returns a `gfx_hal::Backend::Surface`.

omni-viral:
Remove requirement for enabling features transitively to all crates. All common features now
relayed through rendy-util.
Make winit optional.
Refactor backend matching macros.
cargo fmt run.
Fixes #39 #111

Those changes are mostly code move, renames etc.

P.S. Sorry for making this all in one PR

### DIFF
--- a/graph/src/node/present.rs
+++ b/graph/src/node/present.rs
@@ -391,9 +391,21 @@ where
         assert_eq!(images.len(), 1);
 
         let input_image = images.into_iter().next().unwrap();
+        let extent = factory
+            .get_surface_compatibility(&self.surface)
+            .0
+            .current_extent
+            .unwrap_or_else(|| {
+                ctx.get_image(input_image.id)
+                    .expect("Context must contain node's image")
+                    .kind()
+                    .extent()
+                    .into()
+            });
 
         let target = factory.create_target(
             self.surface,
+            extent,
             self.image_count,
             self.present_mode,
             gfx_hal::image::Usage::TRANSFER_DST,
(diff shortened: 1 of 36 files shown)


wsi/src/lib.rs
```diff
@@ -286,21 +300,15 @@ unsafe fn create_swapchain<B: Backend>(
...
-        let (window_width, window_height) = surface
-            .window
-            .get_inner_size()
-            .unwrap()
-            .to_physical(hidpi_factor)
-            .into();
-        gfx_hal::window::Extent2D {
-            width: end.width.min(start.width.max(window_width)),
-            height: end.height.min(start.height.max(window_height)),
+    if let Some(current_extent) = capabilities.current_extent {
```
Frizi: `current_extent` should always take precedence over provided one when available

graph/src/node/present.rs
```diff
@@ -483,8 +495,20 @@ where
             // TODO: use retired swapchains once available in hal and remove that wait
             factory.wait_idle().unwrap();
 
+            let extent = factory
+                .get_surface_compatibility(self.target.surface())
```
Frizi: This call is heavy right now. If the extent provided to `recreate` would be just a fallback, this could always be just an image extent, thus not requring the compatibility query at all.

graph/src/node/present.rs
```diff
@@ -391,9 +391,21 @@ where
         assert_eq!(images.len(), 1);
 
         let input_image = images.into_iter().next().unwrap();
+        let extent = factory
+            .get_surface_compatibility(&self.surface)
```
Frizi: same thing here. Can be avoided

graph/src/node/present.rs
omni-viral: No. In general case you have to provide exactly `current_extent` unless capabilities structure contains `None`

wsi/src/lib.rs
omni-viral: You probably right. Edge case where user intentionally provides different than current extent relying on implementation behavior is edge enough so we could support it only on demand.

graph/src/node/present.rs
Frizi: Yes, but this is irrelevant if `current_extent` is always taken into account instead when available.

Frizi:
bors retry

omni-viral:
bors retry

omni-viral:
bors retry

omni-viral:
bors retry

PR #134, opened 2019-05-19
Move shader-compiler feature to full

Frizi:


omni-viral:
I have similar change in #133
Except you don't include this feature into defaults.

PR #141, opened 2019-05-22
Raw string source shader support

### LINKED ISSUE #136: Shaders cannot be loaded from strings?
It looks like the only built-in way to compile a shader is from a file with a path:  https://github.com/amethyst/rendy/blob/master/shader/src/shaderc.rs#L57

Can we have a way to just have it take `Into<&str>` or something instead, and load the shader source from that string instead of opening the file itself?  That would make it possible for things like `ggez` to use it without jumping through hoops to mess with file paths, and I expect would be necessary for using on the web where you have no file access to begin with.

jaynus:
Deprecates StaticShaderInfo in favor of PathBufShaderInfo. Implements SourceShaderInfo publically which allows for providing raw GLSL source code to pass to shaderc-rs. Adds new example source_shaders for usage. All examples updated for deprecated functionality.

This move soft-deprecates the current StaticShaderInfo which actually wasn't in line with the current implementation; this was supposed to support string shader info to begin with, but instead just provided a path. This instead deprecates that access, and pushes all implementations to PathBufShaderInfo, which enforces a proper PathBuf be supplied (this will actually fix support for non-english user filesystems better).

Then, this changes the current SourceShaderInfo to FileShaderInfo, to better reflect its implementation. Finally, a new implementation SourceCodeShaderInfo is made, which loads from source, and then is aliased as SourceShaderInfo for public consumption.

Closes #136
Closes #140

### DIFF
--- a/rendy/examples/meshes/main.rs
+++ b/rendy/examples/meshes/main.rs
@@ -21,7 +21,7 @@ use {
         memory::Dynamic,
         mesh::{Mesh, Model, PosColorNorm},
         resource::{Buffer, BufferInfo, DescriptorSet, DescriptorSetLayout, Escape, Handle},
-        shader::{ShaderKind, SourceLanguage, SpirvShader, StaticShaderInfo},
+        shader::{PathBufShaderInfo, ShaderKind, SourceLanguage, SourceShaderInfo, SpirvShader},
         wsi::winit::{Event, EventsLoop, WindowBuilder, WindowEvent},
     },
     std::{cmp::min, mem::size_of, time},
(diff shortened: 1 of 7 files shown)


icefoxen:
HAH!  Beat you to it!  :grin:   Your solution is more complete though.

rendy/examples/meshes/main.rs
```diff
@@ -43,15 +43,16 @@ type Backend = rendy::metal::Backend;
 type Backend = rendy::vulkan::Backend;
 
 lazy_static::lazy_static! {
-    static ref VERTEX: SpirvShader = StaticShaderInfo::new(
-        concat!(env!("CARGO_MANIFEST_DIR"), "/examples/meshes/shader.vert"),
+    static ref VERTEX: SpirvShader = SourceShaderInfo::new(
+        unsafe { std::str::from_utf8_unchecked(include_bytes!(concat!(env!("CARGO_MANIFEST_DIR"), "/examples/meshes/shader.vert"))) },
```
omni-viral: How about `include_str!`?

omni-viral: `include_*` macros takes relative path, so you can simply use `"shader.vert"`

shader/src/shaderc.rs
```diff
@@ -89,11 +89,96 @@ where
...
+        stage_from_kind(&self.kind)
+    }
+}
+
 /// Shader info with static data.
-pub type StaticShaderInfo = SourceShaderInfo<&'static str, &'static str>;
+pub type SourceShaderInfo = SourceCodeShaderInfo<&'static str, &'static str, &'static str>;
+
+/// DEPRECATED. USE `PathBufShaderInfo` INSTEAD!
+#[deprecated(since = "2.0", note = "StaticShaderInfo will be removed in favor of PathBufShaderInfo soon. Please move to that implementation.")]
```
omni-viral: Since "2.0"? 😄

rendy/examples/meshes/main.rs
```diff
@@ -43,15 +43,16 @@ type Backend = rendy::metal::Backend;
...
+        concat!(env!("CARGO_MANIFEST_DIR"), "/examples/meshes/shader.vert").into(),
         ShaderKind::Vertex,
         SourceLanguage::GLSL,
         "main",
     ).precompile().unwrap();
 
-    static ref FRAGMENT: SpirvShader = StaticShaderInfo::new(
-        concat!(env!("CARGO_MANIFEST_DIR"), "/examples/meshes/shader.frag"),
+    static ref FRAGMENT: SpirvShader = PathBufShaderInfo::new(
```
omni-viral: So which one examples should use?
Probably `SourceShaderInfo`, that will make them extractable.

jaynus: Swapped all examples to `SourceShaderInfo`

shader/src/shaderc.rs
jaynus: Woops. Thanks :P

rendy/examples/meshes/main.rs
jaynus: Done!

jaynus:
bors retry

jaynus:
bors retry

PR #143, opened 2019-05-23
Add `premultiply_alpha` option to `ImageTextureConfig`

termhn:
Adds an option to have the loader premultiply image alpha when loading. This is desirable in most cases (http://www.adriancourreges.com/blog/2017/05/09/beware-of-transparent-pixels/), however, is left off by default because it could be unexpected default behavior.

This is technically a breaking change, however the number of users that would actually be affected by this is tiny (those using ImageTextureConfig without Default), so it's possibly worth just putting in a patch release.

### DIFF
--- a/texture/src/format/image.rs
+++ b/texture/src/format/image.rs
@@ -104,6 +104,11 @@ pub struct ImageTextureConfig {
     #[derivative(Default(value = "false"))]
     /// Automatically generate mipmaps for this image
     pub generate_mips: bool,
+    #[derivative(Default(value = "false"))]
+    /// Premultiply the alpha channel of the image, if there is one. Note that this
+    /// means an image stored with non-premultiplied alpha will become premultiplied,
+    /// rather than indicating that the supplied image is premultiplied to begin with.
+    pub premultiply_alpha: bool,
 }
 
 #[cfg(feature = "serde")]
@@ -161,6 +166,16 @@ macro_rules! dyn_format {
     }};
 }
 
+/// pixel.channel_count() must be 4
+fn premultiply_alpha<P: image::Pixel<Subpixel = u8>>(pixel: &mut P) 
+{
+    let channels_mut = pixel.channels_mut();
+    let alpha = channels_mut[3] as f32 / 255.0;
+    channels_mut[0] = (channels_mut[0] as f32 * alpha).min(255.0).max(0.0) as u8;
+    channels_mut[1] = (channels_mut[1] as f32 * alpha).min(255.0).max(0.0) as u8;
+    channels_mut[2] = (channels_mut[2] as f32 * alpha).min(255.0).max(0.0) as u8;
+}
+
 /// Attempts to load a Texture from an image.
 pub fn load_from_image<R>(
     mut reader: R,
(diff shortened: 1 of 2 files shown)


omni-viral:
I think `Texture` instance should indicate if image is alpha-premultiplied

termhn:
> I think `Texture` instance should indicate if image is alpha-premultiplied

I think this is an implementation detail in some way, though I'm not opposed to adding this as a field necessarily. Just seems like something that a consumer of `rendy` would decide to either use or not use rather than mix and match, so tagging every `Texture` instance seems a bit wasteful?

> Maybe also bump the version so we can get rid of the git dependency in amethyst/amethyst#1637 ?

@torkleyy Yea for sure... on this subject, @omni-viral do you think we should bump patch or minor version for this? I'm kinda inclined for patch version personally even though it's technically breaking, but I guess it's really not that big of a deal to just have a new minor version?

torkleyy:
> I'm kinda inclined for patch version personally even though it's technically breaking, but I guess it's really not that big of a deal to just have a new minor version?

Please do release a minor. It's better to release an extra version just to be safe; there's no harm with releasing many minor versions. What is bad is when we have dependencies accidentally breaking another library, causing breakage in a whole range of other libraries. That produces

1) issues / PRs for all the libraries depending on the broken dependency
2) potentially long waiting times for passively maintained crates
3) quick fixes to lock dependencies
4) people not trusting semver, thus locking the exact version in Cargo.toml, leading to larger binaries + version incompatibilities

Releasing a new version is more opt-in, in that it only requires action when the additional feature is needed, and not when accidental / minor breakage occurred.

texture/src/format/image.rs
```diff
@@ -104,6 +104,11 @@ pub struct ImageTextureConfig {
     #[derivative(Default(value = "false"))]
     /// Automatically generate mipmaps for this image
     pub generate_mips: bool,
+    #[derivative(Default(value = "false"))]
+    /// Premultiply the alpha channel of the image, if there is one. Note that this
+    /// means an image stored with non-premultiplied alpha will become premultiplied,
+    /// rather than indicating that the supplied image is premultiplied to begin with.
+    pub premultiply_alpha: bool,
```
torkleyy: What you can do to minimize minor bumps is to add a private field with a `serde(skip)` attribute.

The disadvantage of that is that you need to create constructors, and `...Default::default()` cannot be used.

Frizi: We use that struct directly as serialized options in amethyst. If you anticipate that this might be changed frequently, then indeed this should use a builder pattern, and amethyst should have it's own serialized format that can be converted into that.

texture/src/format/image.rs
```diff
@@ -216,21 +231,35 @@ where
...
+                        }
+                    }
+                    (
+                        img.into_vec(),
+                        dyn_format!(Rgba, _8, config.repr),
+                        Swizzle::NO,
+                    )
+                },
                 DynamicImage::ImageBgr8(img) => (
                     img.into_vec(),
```
Frizi: `Bgr` could use premultiply alpha as well.

Frizi:
For completeness, `ImageLumaA8` should use `premultiply_alpha` option as well.

omni-viral:
@termhn If original image had premultiplied alpha or if during texture creation rendy premultiplies alpha then result is exactly the same - user must use `Texture` in environment that expects premultiplied alpha. That's why `Texture` should indicate actual content state.

texture/src/format/image.rs
```diff
@@ -283,4 +290,5 @@ where
         .with_kind(kind)
         .with_view_kind(config.kind.view_kind())
         .with_sampler_info(config.sampler_info))
+        .with_premultiplied_alpha(config.premultiply_alpha)
```
Frizi: I think we need 3 or 4 enum variants for `premultiply_alpha` here, as you might as well load premultiplied images from storage without premultiplying at runtime. In fact, reverse could be true as well to some degree maybe?

So, we might have "Premultiplied" and "NotPremultiplied" both at input and output. The final texture should signal the output state (what texture ends up being after loading), and runtime action should be taken when input differs from output.

Does that make any sense? :P

termhn: IMO rendy itself should not care about this, and this is why I was skeptical of even having a premultiplied field on textures. Whether an image's alpha is premultiplied or not should be something that is handled *by the user of rendy* not by rendy itself. Unless we want to add utilities to premultiply things on the fly based on user configuration, but I don't think we do. Adding it only in the `image` feature makes sense because it is where we provide the ability to load things from disk. For the rest of rendy, we're just operating on raw data, and the gpu doesn't actually care if the image's alpha is premultiplied or not, only the user's algorithms do.

termhn: Basically, my point is that at this point, the field on a Texture is only there to signal that "this image *IS CURRENTLY* in a premultiplied state", not to signify whether it "*should*" or "*should not*" be premultiplied.

omni-viral: I guess if we would find ourselves in situation where we would want ask `rendy` if the texture has pre-multiplied alpha or not then we can think of adding parameters to `Texture`.

omni-viral:
bors delegate+

texture/src/format/image.rs
Frizi: @omni-viral in that case, why should we even have that field then? If you don't plan to query it, it's useless.

PR #146, opened 2019-05-24
Add frames in flight to graph context

swooster:
Adds `.frames_in_flight` to `GraphContext` so nodes can configure themselves to use the right quantity of resources without needing to be coupled to `aux`.

Also simplifies the `meshes` example to directly use `Scene` as the `aux` type, now that `frames` doesn't need to be indirectly passed through `aux`.

### DIFF
--- a/graph/src/graph/mod.rs
+++ b/graph/src/graph/mod.rs
@@ -40,6 +40,8 @@ device_owned!(Graph<B, T: ?Sized>);
 pub struct GraphContext<B: Backend> {
     buffers: Vec<Option<Handle<Buffer<B>>>>,
     images: Vec<Option<(Handle<Image<B>>, Option<gfx_hal::command::ClearValue>)>>,
+    /// Number of potential frames in flight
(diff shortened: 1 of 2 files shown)


swooster:
I was sorely tempted to make a `frames_in_flight` public member, but it seems like most other things use getters, so I've defaulted to doing that. (if you'd prefer otherwise, I'd be happy to change it)

termhn:
Nice! I am conflicted on what I think the best accessor method would be. I would personally like either just a public `frames_in_flight` member, or else a getter that is just `frames_in_flight()` rather than having the `get_` prefix

swooster:
Those are my preferences as well... I'm under the impression Rust leans towards public members where feasible (and I find `.frames_in_flight` cleaner looking), and the graph context is passed via a read-only ref so it's not like it can be altered by a node.

swooster:
Ok, I've gone ahead and pushed a version with a public `frames_in_flight` member.

Issue #142, opened 2019-05-23
Size of push constants are not checked against layouts?


icefoxen:
A `Layout` takes a vec of push constant descriptions, which include the size of the push constant.  However, it appears that no check is done to make sure the size of the data given to `encoder.push_constants()` is actually this size, and the validation layer didn't catch it for me either.  (It also didn't catch that the size of the data I was trying to give it was several times larger than `maxPushConstantsSize`.)  Do we want to add a debug-mode-only bounds check for this?

omni-viral:
This is not invalid usage, except maybe providing more than `maxPushConstantsSize`

omni-viral:
As you may noticed `encoder.push_constants` takes raw `gfx-hal`'s `PipelineLayout` and rendy can't query for it's structure.
We would have to write `PipelineLayout` wrapper that saves `CreateInfo` for us to inspect.

icefoxen:
Oh shoot, you're right.  The info I want is in `rendy::graph::render::Layout`.  Let me look more at how that is related to `PipelineLayout`, if at all.

icefoxen:
For my own reference: `SimpleGraphicsPipelineDesc::layout()` is called in the `RenderGroupDesc::build()` impl for it, which calls [`gfx-hal`'s `Device::create_pipeline_layout'](https://docs.rs/gfx-hal/0.2.0/gfx_hal/device/trait.Device.html#tymethod.create_pipeline_layout).  So yes, unless we store the information elsewhere and/or make a way for the `Draw` method to get its hands on the original `Layout`, we can't do this in rendy itself.

Vulkan's whole point is that it can assume that the data layout you give it matches the data you give it, and doesn't offer many ways to enforce that.  It's up to the user to make sure that's correct.  And if rendy is to remain a thin layer over `gfx-hal` we don't want to go through a whole lot of work enforcing that the data matches what it should; we can do it when it's easy but adding a whole bunch of wrappers for bounds-checking and stuff doesn't sound like what we want.

I'll keep thinking about this and maybe try to come up with an easy way to do the bounds-checking in my own code.  Up to you if you want to close this issue.

malobre:
Maybe this should be closed ?

omni-viral:
I guess so

PR #147, opened 2019-05-24
Mark unsafe functions as unsafe.

icefoxen:
Basically I looked at #142 and said "wait, if passing the wrong parameters to this function can cause memory corruption, and there's no good way we can automatically add checks for that, this function should be unsafe".

So I added `unsafe` to everywhere in `EncoderCommon` that it looked like it needed it, generally 'cause it called an unsafe `gfx-hal` function and passed parameters through without doing any bounds checking.  Then I added unsafe blocks where necessary to make the thing compile.  This PR is offered as a matter for discussion and feedback to figure out how we want to handle this issue; the code is far from complete.

### DIFF
--- a/command/src/buffer/encoder.rs
+++ b/command/src/buffer/encoder.rs
@@ -87,7 +87,7 @@ where
     /// Note that `draw*` commands available only inside renderpass.
     ///
     /// [`draw_indexed`]: ../struct.RenderPassEncoder.html#method.draw_indexed
-    pub fn bind_index_buffer<'b>(
+    pub unsafe fn bind_index_buffer<'b>(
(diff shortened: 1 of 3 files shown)


omni-viral:
`unsafe` functions should have `# Safety` documentation block that explains how it should be used to make things safe.

icefoxen:
Excellent point, thank you.  I've added it for a bunch of methods.  Others have placeholders I'll fix when I have energy to think about it.

omni-viral:
You can add a link to vulkan spec page for the method instead of long explanations.

omni-viral:
Followed by #165 165

PR #150, opened 2019-05-26
Don't panic when freeing resources if already panicking.

azriel91:
When using `rendy` in a test application, and the application panics, sometimes there are resources that are still held when `DynamicAllocator::dispose(self)` is called, causing an illegal instruction (segfault) on Windows.

This PR prevents a double panic, though I don't understand `rendy` enough to tell how to free up the resources before the allocator.

### DIFF
--- a/memory/src/allocator/dynamic.rs
+++ b/memory/src/allocator/dynamic.rs
@@ -2,6 +2,7 @@ use std::{
     collections::{BTreeSet, HashMap},
     ops::Range,
     ptr::NonNull,
+    thread,
 };
 
 use {
@@ -473,8 +474,16 @@ where
 
     /// Perform full cleanup of the memory allocated.
     pub fn dispose(self) {
-        for (index, size) in self.sizes {
-            assert_eq!(size.chunks.len(), 0, "SizeEntry({}) is still used", index);
+        if !thread::panicking() {
+            for (index, size) in self.sizes {
+                assert_eq!(size.chunks.len(), 0, "SizeEntry({}) is still used", index);
+            }
+        } else {
+            for (index, size) in self.sizes {
+                if size.chunks.len() != 0 {
+                    log::error!("Memory leak: SizeEntry({}) is still used", index);
+                }
+            }
         }
     }
 }


azriel91:
By the way, I managed to get this when playing around with `rendy`:

```markdown
...
[ERROR][rendy_resource::escape] Terminal must be dropped after all `Escape`s
[ERROR][rendy_resource::escape] Terminal must be dropped after all `Escape`s
[ERROR][rendy_memory::allocator::dynamic] Memory leak: SizeEntry(8192) is still used
[ERROR][rendy_memory::allocator::dynamic] Memory leak: SizeEntry(1024) is still used
[ERROR][rendy_memory::allocator::dynamic] Memory leak: SizeEntry(65536) is still used
[ERROR][rendy_memory::allocator::dynamic] Memory leak: SizeEntry(524288) is still used
[ERROR][rendy_memory::allocator::dynamic] Memory leak: SizeEntry(4194304) is still used
[ERROR][rendy_resource::escape] `Escape` was dropped after a `Terminal`?
...
```

azriel91:
More info on the segfault, given:

* an empty render graph (`GraphBuilder::new()`)
* 100 test functions

The segfault does not happen if I run with `cargo test -- --test-threads 1`. It usually does if you have multiple threads.

```bash
git clone git@github.com:azriel91/amethyst_rendy_test.git
cd amethyst_rendy_test
cargo update # in case you already had it
cargo test -- --nocapture
# should segfault
```

This indicates that the test threads are sharing a common memory space (see previous comment logs). Theoretically having 1 process per `rendy` wouldn't have them share the memory space, but that's not a realistic viable solution (writing another test executor, and runtime cost to spawn process).

Frizi:
This one
```
[ERROR][rendy_resource::escape] `Escape` was dropped after a `Terminal`?
```
is basically about releasing resource after the allocator. In the end, it means there is no leak, just the drop order is wrong somewhere. You can put a breakpoint on that error and inspect the stack to see what is being deallocated, then make sure to deallocate it before factory is dropped if possible.

azriel91:
bors retry

azriel91:
bors retry

omni-viral:
bors retry

PR #152, opened 2019-05-27
Avoid unnecessary mutable borrows in examples

swooster:
In examples, switches `&mut aux` to `&aux` to match signatures of `Graph` methods.


swooster:
I initially had the mistaken impression that graph nodes might somehow be able to interact with `aux` because of the triangle example's use of `&mut ()`. I figure making the references read-only could help the next newbie.

omni-viral:
Thanks. `Graph` originally took `aux: &mut T` argument. Those are remnants.

bors r+

omni-viral:
bors retry

PR #153, opened 2019-05-30
Add more ways to construct DescriptorRanges

kvark:
With the current API, the use has to have a slice of bindings in order to create the subject struct. Since this isn't always the case, it forces unnecessary friction on the caller side, while we can have a more flexible API with iterators and/or ability to add individual bindings.


Frizi:
bors retry

kvark:
OSX times out

omni-viral:
bors retry

PR #156, opened 2019-06-03
Fix another rendy-memory assert

rukai:
I hit another very similar looking assert.
Please be aware that I don't properly understand this and I just cargo-culted your library so that my program works. :P

I also changed `blocks` from a u32 to a u64, because it looks like high_bit is supposed to get shifted all the way. But this didn't affect my program either way.

Also also, it looks like there are even more similar asserts lying around :eyes: But I have no idea if they are correct or not.

### DIFF
--- a/memory/src/allocator/dynamic.rs
+++ b/memory/src/allocator/dynamic.rs
@@ -588,8 +588,7 @@ where
 
     /// Check if there are free blocks.
     fn is_unused(&self, block_size: u64) -> bool {
-        let blocks = (self.size() / block_size) as u32;
-        debug_assert!(blocks <= MAX_BLOCKS_PER_CHUNK);
+        let blocks = (self.size() / block_size).min(MAX_BLOCKS_PER_CHUNK as u64);
 
         let high_bit = 1 << (blocks - 1);
         let mask = (high_bit - 1) | high_bit;


memory/src/allocator/dynamic.rs
```diff
@@ -588,8 +588,7 @@ where
 
     /// Check if there are free blocks.
     fn is_unused(&self, block_size: u64) -> bool {
-        let blocks = (self.size() / block_size) as u32;
-        debug_assert!(blocks <= MAX_BLOCKS_PER_CHUNK);
```
omni-viral: Right. This is basically same assert that can be failed due to overallocation.
While changing type of `blocks` is correct and consistent with other code (although current implementation ensures it is never > 128), `high_bit` type is inferred from `self.blocks` which is `u64`.

omni-viral:
Assert in `from_memory` function should hold as there is no reason dedicated allocation will overallocate.

memory/src/allocator/dynamic.rs
rukai: Cool, thanks for the explanation.

rukai:
Bors retry

omni-viral:
bors retry

azriel91:
bors retry

azriel91:
bors retry
(mac nightly build too slow)

PR #161, opened 2019-06-10
Reduce logging level

torkleyy:


factory/src/factory.rs
```diff
@@ -1120,7 +1120,7 @@ where
     let heaps = heaps.into_iter().collect::<SmallVec<[_; 16]>>();
     let types = types.into_iter().collect::<SmallVec<[_; 32]>>();
 
-    log::info!("Heaps: {:#?}\nTypes: {:#?}", heaps, types);
+    log::trace!("Heaps: {:#?}\nTypes: {:#?}", heaps, types);
```
Frizi: If queues and above are debug, i think this should be as well. It's one time so not a big deal

rendy/examples/quads/main.rs
```diff
@@ -512,14 +512,14 @@ where
...
-            log::info!("Acquire {:?} : {:#?}", stages, barriers);
+            log::trace!("Acquire {:?} : {:#?}", stages, barriers);
             encoder.pipeline_barrier(stages, hal::memory::Dependencies::empty(), barriers);
         }
         encoder.dispatch(QUADS, 1, 1);
 
         {
             let (stages, barriers) = gfx_release_barriers(ctx, &*buffers, None);
-            log::info!("Release {:?} : {:#?}", stages, barriers);
+            log::trace!("Release {:?} : {:#?}", stages, barriers);
```
Frizi: i think we actually want `info`s here, this is an example code

torkleyy:
Resolved, would be great if we could get this into a new patch version.

torkleyy:
bors r=Frizi

torkleyy:
Seriously Travis, why?
bors retry

PR #162, opened 2019-06-11
Build full feature on docs.rs

torkleyy:


rendy/Cargo.toml
```diff
@@ -97,4 +97,7 @@ required-features = ["base", "wsi-winit"]
 
 [[example]]
 name = "source_shaders"
-required-features = ["base", "wsi-winit"]
\ No newline at end of file
+required-features = ["base", "wsi-winit"]
+
+[package.metadata.docs.rs]
+features = ["base", "mesh-obj", "serde-1", "texture-image", "texture-palette", "spirv-reflection", "wsi-winit"]
```
omni-viral: You intentionally ommited `spirv-compilation`?
Otherwise just specify `full`

torkleyy: Yes, because I think it would need additional dependencies, right?

torkleyy: ninja I believe

Frizi: Ninja is only required on windows afaik

torkleyy: Ah right, and docs.rs only builts for Linux by default!

torkleyy:
```suggestion
features = ["full"]
```

PR #164, opened 2019-06-14
Allow render pass to use surface as an attachment.

omni-viral:
Remove need in dedicated presentation node when final image is rendered by single render pass.

### DIFF
--- a/graph/src/node/render/pass.rs
+++ b/graph/src/node/render/pass.rs
@@ -370,7 +370,7 @@ where
         attachments.sort();
         attachments.dedup();
 
-        let find_attachment_node_image = |id: ImageId| {
+        let find_attachment_node_image = |id: ImageId| -> &NodeImage {
             images
                 .iter()
                 .find(|a| a.id == id)
@@ -522,7 +522,14 @@ where
                         .map(|&i| {
                             (
                                 attachments.iter().position(|&a| a == i).unwrap(),
-                                gfx_hal::image::Layout::General,
+                                match i {
+                                    Either::Left(image_id) => {
+                                        find_attachment_node_image(image_id).layout
+                                    }
+                                    Either::Right(RenderPassSurface) => {
+                                        gfx_hal::image::Layout::ShaderReadOnlyOptimal
+                                    }
+                                },
                             )
                         })
                         .collect(),
@@ -532,14 +539,28 @@ where
                         .map(|&c| {
                             (
                                 attachments.iter().position(|&a| a == c).unwrap(),
-                                gfx_hal::image::Layout::ColorAttachmentOptimal,
+                                match c {
+                                    Either::Left(image_id) => {
+                                        find_attachment_node_image(image_id).layout
+                                    }
+                                    Either::Right(RenderPassSurface) => {
+                                        gfx_hal::image::Layout::ColorAttachmentOptimal
+                                    }
+                                },
                             )
                         })
                         .collect(),
                     depth_stencil: subpass.depth_stencil.map(|ds| {
                         (
                             attachments.iter().position(|&a| a == ds).unwrap(),
-                            gfx_hal::image::Layout::General,
+                            match ds {
+                                Either::Left(image_id) => {
+                                    find_attachment_node_image(image_id).layout
+                                }
+                                Either::Right(RenderPassSurface) => {
+                                    gfx_hal::image::Layout::DepthStencilAttachmentOptimal
+                                }
+                            },
                         )
                     }),
                 })
--- a/graph/src/node/render/pass.rs
+++ b/graph/src/node/render/pass.rs
@@ -16,20 +16,27 @@ use {
             render::group::{RenderGroup, RenderGroupBuilder},
             BufferAccess, DynNode, ImageAccess, NodeBuffer, NodeBuilder, NodeImage,
         },
+        wsi::{Surface, Target},
         BufferId, ImageId, NodeId,
     },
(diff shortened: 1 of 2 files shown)


graph/src/node/render/pass.rs
```diff
@@ -453,28 +621,12 @@ where
...
-                        &*attachment
-                    })),
-            );
+            let (stages, barriers) = gfx_acquire_barriers(ctx, &buffers, &images);
 
             if !barriers.is_empty() {
                 let initial = command_pool.allocate_buffers(1).pop().unwrap();
                 let mut recording = initial.begin(MultiShot(SimultaneousUse), ());
-                log::trace!("Acquire {:?} : {:#?}", stages, barriers);
+                log::debug!("Acquire {:?} : {:#?}", stages, barriers);
```
Frizi: why is that log level changed?

graph/src/node/render/pass.rs
```diff
@@ -370,32 +519,34 @@ where
                     inputs: subpass
                         .inputs
                         .iter()
-                        .map(|&id| {
+                        .map(|&i| {
                             (
-                                attachment_ids.iter().position(|&a| a == id).unwrap(),
-                                find_attachment(id).layout,
+                                attachments.iter().position(|&a| a == i).unwrap(),
+                                gfx_hal::image::Layout::General,
```
Frizi: Is that `General` hardcode correct here? Same for depth.

graph/src/node/render/pass.rs
```diff
@@ -579,28 +730,74 @@ where
...
-            command_pool,
-            command_cirque,
-
-            acquire,
-            release,
-
-            relevant: relevant::Relevant,
+        let node: Box<dyn DynNode<B, T>> = match node_target {
+            Some(target) => {
+                log::debug!("Construct RenderPassNodeWithSurface");
```
Frizi: this should be `trace`

graph/src/node/render/pass.rs
```diff
@@ -579,28 +730,74 @@ where
...
+                            release: factory.create_semaphore().unwrap(),
+                            index: 0,
+                        })
+                        .collect(),
+                    free_acquire: factory.create_semaphore().unwrap(),
+                    target,
+                })
+            }
+            None => {
+                log::debug!("Construct RenderPassNodeWithoutSurface");
```
Frizi: `trace`

graph/src/node/render/pass.rs
```diff
@@ -649,7 +844,66 @@ pub struct RenderPassNode<B: Backend, T: ?Sized> {
...
+struct PerImage<B: Backend> {
+    framebuffer: B::Framebuffer,
+    acquire: B::Semaphore,
+    release: B::Semaphore,
+    index: usize,
+}
+
+#[derive(derivative::Derivative)]
+#[derivative(Debug(bound = ""))]
+struct RenderPassNodeWithSurface<B: Backend, T: ?Sized> {
```
Frizi: `RenderPassNode` was previously `pub`. I think now this and `RenderPassNodeWithoutSurface` should be `pub` to match that. Was it actually correct before in being public?

graph/src/node/render/pass.rs
```diff
@@ -665,45 +919,232 @@ where
...
-                                    queue.id(),
+            if let Some(next) = &next {
+                let ref mut for_image = per_image[next[0] as usize];
+
+                let force_record =
+                    subpasses
+                        .iter_mut()
+                        .enumerate()
+                        .any(|(subpass_index, subpass)| {
+                            subpass.groups.iter_mut().any(|group| {
```
Frizi: that `any` is not correct, as it will short circuit and not call `prepare` some all groups. It was `fold` before for a reason.

graph/src/node/render/pass.rs
```diff
@@ -665,45 +919,232 @@ where
...
+        } = self;
+
+        let submit = command_cirque.encode(frames, command_pool, |mut cbuf| {
+            let index = cbuf.index();
+
+            let force_record = subpasses
+                .iter_mut()
+                .enumerate()
+                .any(|(subpass_index, subpass)| {
+                    subpass.groups.iter_mut().any(|group| {
```
Frizi: Again, this has to be `fold` that avoids short-circuiting.

graph/src/node/render/pass.rs
omni-viral: Not really. Should be fixed now.

graph/src/node/render/pass.rs
omni-viral: Builder returns is as `dyn DynNode` anyway. So there is no way to get instance of type `RenderPassNodeWithSurface`

graph/src/node/render/pass.rs
omni-viral: You're absolutely right.

graph/src/node/render/pass.rs
Frizi: Ok, let's just keep it private then.

graph/src/node/render/pass.rs
omni-viral: Just think that something with meaningful data can be `debug` and simple "that happens" - `trace`.

Frizi: As long as it's not logging every frame, i think we should be ok. It was recently changed to traces though. Amethyst was way too verbose before.

termhn:
rip timeout

bors r+

PR #166, opened 2019-06-19
Update status

kellytk:


omni-viral:
bor r-

Also `wgpu-rs` and `WebRender` fork are using parts of `rendy`.
Could you add that too?

kellytk:
@omni-viral Where shall I link to to include mention of WebRender?

omni-viral:
Let's wait 'til `WebRender` will use `rendy` in master.

bors r+

PR #173, opened 2019-07-09
fix potentially unsound Factory::drop

lcnr:
As far as I know calling dispose requires unique access on a device.  This is guaranteed by calling `device.wait_idle()`. As we previously discarded the result of this call, it may have silently failed. This allowed us to cause UB by disposing currently used resources later on.

I am currently not that knowledgeable about both vulkan and rendy, so please correct me if anything I said was wrong.

### DIFF
--- a/factory/src/factory.rs
+++ b/factory/src/factory.rs
@@ -110,7 +110,10 @@ where
 {
     fn drop(&mut self) {
         log::debug!("Dropping factory");
-        let _ = self.wait_idle();
+        match self.wait_idle() {
+            Err(HostExecutionError::DeviceLost) | Ok(()) => (),
+            Err(err) => panic!("{}", err),
+        }
 
         unsafe {
             // Device is idle.


omni-viral:
Thanks!
I think we can proceed on `DeviceLost` error though and panic only on OOM

lcnr:
idk, maybe :man_shrugging:
I don't understand most of this yet.

omni-viral:
This code is correct, but it reads wrong.
It's as if you were saying that for this function to return any other error is a bug and you're checking this invariant. Would you kindly replace with match and perform `panic!` with appropriate message?

lcnr:
@omni-viral still not perfectly satisfied with my code. Is my reasoning for why DeviceLost is allowed correct?

omni-viral:
Quote from Vulkan spec here: https://www.khronos.org/registry/vulkan/specs/1.0-extensions/html/vkspec.html#devsandqueues-lost-device

"""
For any command that may return VK_ERROR_DEVICE_LOST, for the purpose of determining whether a command buffer is in the pending state, or whether resources are considered in-use by the device, a return value of VK_ERROR_DEVICE_LOST is equivalent to VK_SUCCESS.
"""

factory/src/factory.rs
```diff
@@ -110,7 +110,12 @@ where
 {
     fn drop(&mut self) {
         log::debug!("Dropping factory");
-        let _ = self.wait_idle();
+        match self.wait_idle() {
+            Err(HostExecutionError::OutOfDeviceMemory) => panic!("Out of device memory"),
```
omni-viral: For any other errors (in case new will be added) just match as `Err(err)` and panic using `Display` implementation for the error.

omni-viral:
bors retry

PR #174, opened 2019-07-09
[WIP] Basic update to winit 0.20.0-alpha1

icefoxen:
This version of winit changes how event loops are handled, so
I wanted to experiment and see how it works.  Seems like it
should be compatible, though disposing of things might get a
little squirrelly; haven't tried that yet.  Examples still need
to be updated, for now `poll_events()` is just commented out so
they build.

I optimistically intend to keep this branch more or less up to date
as rendy and winit carry on, and then make it non-WIP when winit 0.20
stabilizes.

### DIFF
--- a/factory/src/factory.rs
+++ b/factory/src/factory.rs
@@ -1154,7 +1154,7 @@ rendy_wsi::with_winit! {
         B: Backend,
     {
         /// Create rendering surface from window.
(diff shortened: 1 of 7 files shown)


omni-viral:
Can you agree with #175 ?

omni-viral:
Now #235 is merged and winit isn't even used by rendy. Except in examples.

PR #175, opened 2019-07-10
Optional support for using winit 0.20. Opt-in using feature "winit-20".

aclysma:
This would allow optional usage of winit 0.20.0. Examples and default build behavior would continue to use earlier winit versions.

### DIFF
--- a/factory/src/factory.rs
+++ b/factory/src/factory.rs
@@ -1154,7 +1154,7 @@ rendy_wsi::with_winit! {
         B: Backend,
     {
         /// Create rendering surface from window.
-        pub fn create_surface(&mut self, window: &rendy_wsi::winit::Window) -> Surface<B> {
+        pub fn create_surface(&mut self, window: &rendy_wsi::WinitWindow) -> Surface<B> {
             profile_scope!("create_surface");
             Surface::new(&self.instance, window)
         }
--- a/rendy/Cargo.toml
+++ b/rendy/Cargo.toml
@@ -13,6 +13,7 @@ readme = "../README.md"
 
 [features]
 serde-1 = ["rendy-factory/serde-1", "rendy-mesh/serde-1", "rendy-texture/serde-1", "rendy-shader/serde-1", "rendy-util/serde-1"]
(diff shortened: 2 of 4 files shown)


omni-viral:
Can you agree with #174 ?

icefoxen:
Ahahaha, great minds obviously think alike, hm?  I'm not convinced that having a feature flag to select which version of winit is the way to go, but it honestly kinda isn't the worst solution to The Dependency Problem.  But solving that problem is, well, what `wsi` is there for already...

aclysma:
I think you’ll eventually want icefoxen’s, but it drops support for winit .19. I think this PR and icefoxen’s PR can be mixed and matched depending on the timeframe you want to add support for .20, drop support for .19, and move the examples over to .20. These three things don’t necessarily have to happen at the same time and the timing is a matter of taste. 

Supporting both simultaneously doesn’t make too much of a mess in the library, but it might be tricky to do cleanly in the examples.

I would suggest supporting both in the library until winit .20 is stable and the library ecosystem has mostly moved over to using it. 

To do this, you could take this PR now. Then when you’re ready to move the examples to .20, take icefoxen’s and make .20 default and .19 opt-in, then drop .19. Or if you feel ready to move over fully now and drop support for .19, you could skip this PR.

icefoxen:
Yeah I agree with all the above. So we need a policy decision, @omni-viral .  Do you want to:

 * Have support for both winit 0.19 and 0.20 via feature flags?  0.20 is currently alpha, but this would let you start messing with it in `master` NOW, without needing to update as much later.  But like @aclysma said, making the examples play nice with this is trickier.  If so, take this PR, and I'll make sure rendy stays up to date if winit 0.20 changes.
 * Only support winit 0.19, and once 0.20 is released switch over to it and drop support for 0.19?  If so, close this one and take #174 once it's done.

omni-viral:
@aclysma As we can't remove features in releases, we will not release with this feature-versioning. And I was going to release soon (tm).
If you personally want to work with winit 0.20 I see no reason to not use a patch in your workspace.

aclysma:
No worries, sounds good!

PR #181, opened 2019-08-04
Update shaderc version to 0.6

barskern:
This update enables shaderc to choose either shared or static versions
of the SPIRV-Tools, depending on what is installed on the system.

This is a minor update because Python2 is no longer supported ([relevant commit]).

[relevant commit]:
https://github.com/google/shaderc-rs/commit/4689e504ef62a05b8096356e3ad1b025b12283d4

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,13 +15,17 @@ matrix:
     - env: MACOSX_DEPLOYMENT_TARGET=10.9
       os: osx
       osx_image: xcode10.2
-      install: travis_wait 30 cargo build --verbose
+      # We need travis_wait here to ensure the build doesn't timeout due to
+      # shader compilation
+      script: travis_wait 30 make
 
     # iPhoneOS 64bit
     - env: TARGET=aarch64-apple-ios
       os: osx
       osx_image: xcode10.2
-      install: travis_wait 30 cargo build --verbose
+      # We need travis_wait here to ensure the build doesn't timeout due to
+      # shader compilation
+      script: travis_wait 30 make
 
     # Windows 64bit
     - os: windows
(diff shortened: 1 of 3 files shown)


barskern:
I am not able to set the version of Python for Travis to be 3.*. I cannot seem to find out how to set the version without also setting the language to Python.

barskern:
This currently works on windows and linux now, however the mac build seems to time out. I'm not sure how to make it run faster.

barskern:
Woot! Finally made this compile!

.travis.yml
```diff
@@ -14,15 +14,18 @@ matrix:
     # macOS 64bit
     - env: MACOSX_DEPLOYMENT_TARGET=10.9
       os: osx
-      osx_image: xcode9
+      osx_image: xcode10.2
+      install: travis_wait 30 cargo build --verbose
```
omni-viral: You build crate in `install`? Is that right thing to do?

barskern: Ah that is a good point. It should actually probably rather be `script: travis_wait 30 make`. I will try that when I'm back. 

Further we should probably also set the default of `install` to nothing, as by default it is `cargo build --verbose`. See [here](https://docs.travis-ci.com/user/languages/rust/).

barskern:
I will squash these if the build is successful. :smile:

omni-viral:
Won't this run `make` twice?

barskern:
No I am pretty sure that when we specify `script` in a build job it overrides the globally specified `script`.

barskern:
Is there anything blocking this PR? It would greatly improve my experience on Archlinux because by default only shared versions of `SPIRV-Tools` get installed, hence it is a bit of a hassle to get the static versions. @omni-viral

omni-viral:
@barskern No, I just got distracted. Can you please fix merge conflict and I will see if we can merge this PR :)

barskern:
@omni-viral Ah okay I see. No worries. Just pushed it now. :smile:

barskern:
There now is ready @omni-viral.

omni-viral:
bors retry

PR #185, opened 2019-08-09
Update all dependencies

rukai:
closes https://github.com/amethyst/rendy/issues/184

PR #186, opened 2019-08-10
Removed size check, add x11 dependency for gfx-backend-vulkan

nodef0:
Tried to run the examples in a x11-vulkan setup, and found that the update to gfx-rs 0.3 now needs an explicit "x11" feature. 
Also, since the change of the `spirv` parameter from `Vec<u8>` to `Vec<u32>` in merge (#185) the size check is not correct.

### DIFF
--- a/util/Cargo.toml
+++ b/util/Cargo.toml
@@ -29,7 +29,7 @@ gfx-hal = "0.3"
 gfx-backend-empty = { version = "0.3", features = ["winit"], optional = true }
 gfx-backend-dx12 = { version = "0.3", features = ["winit"], optional = true }
 gfx-backend-metal = { version = "0.3", features = ["winit"], optional = true }
-gfx-backend-vulkan = { version = "0.3", features = ["winit"], optional = true }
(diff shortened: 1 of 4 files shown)


util/Cargo.toml
```diff
@@ -29,7 +29,7 @@ gfx-hal = "0.3"
 gfx-backend-empty = { version = "0.3", features = ["winit"], optional = true }
 gfx-backend-dx12 = { version = "0.3", features = ["winit"], optional = true }
 gfx-backend-metal = { version = "0.3", features = ["winit"], optional = true }
-gfx-backend-vulkan = { version = "0.3", features = ["winit"], optional = true }
+gfx-backend-vulkan = { version = "0.3", features = ["winit", "x11"], optional = true }
```
omni-viral: We can't just enable every feature in gfx, but reexport it instead.

nodef0: I'm not exactly sure what is the correct way to do that.
I've added a feature 'vulkan-x11', since I don't know how to make 'x11' orthogonal to the selected back-end.

omni-viral:
Should be ok.

bors r+

PR #187, opened 2019-08-12
GL [divide et impera]

omni-viral:


chain/src/collect.rs
```diff
@@ -37,7 +37,7 @@ struct Fitness {
 
 struct ResolvedNode {
     id: usize,
-    family: gfx_hal::queue::QueueFamilyId,
+    family: rendy_core::hal::queue::QueueFamilyId,
```
Frizi: can we at least `use rendy_core::hal;` everywhere? Having those extra long paths is disturbing now.

omni-viral: Sure. It was just auto-replace.

examples/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+        event_loop: EventLoop<()>,
+    }
+
+    enum Never {}
+
+    impl<E> WithAnyWindowedRendy for ExampleWithAnyRendy<E>
+    where
+        E: Example,
+    {
+        type Output = Never;
```
Frizi: This should be `!`

examples/src/meshes/main.rs
```diff
@@ -0,0 +1,544 @@
...
+#[cfg(feature = "spirv-reflection")]
+use rendy::shader::SpirvReflection;
+
+#[cfg(not(feature = "spirv-reflection"))]
+use rendy::mesh::AsVertex;
+
+lazy_static::lazy_static! {
+    static ref VERTEX: SpirvShader = SpirvShader::new(
+        unsafe {
+            let bytes = include_bytes!("vert.spv");
```
Frizi: We need to get `from_bytes` (#213) merged and then it can be used here and in other shaders.

factory/src/config.rs
```diff
@@ -165,27 +164,29 @@ unsafe impl HeapsConfigure for BasicHeapsConfigure {
...
-                            (properties.memory_heaps[mt.heap_index] / 4096).next_power_of_two(),
-                        ),
-                        min_device_allocation: _1mb
-                            .min(properties.memory_heaps[mt.heap_index] / 1048)
-                            .next_power_of_two(),
-                        max_chunk_size: _32mb.min(
-                            (properties.memory_heaps[mt.heap_index] / 128).next_power_of_two(),
-                        ),
-                    }),
+                    // linear: if mt
```
Frizi: commneted out code

graph/src/node/mod.rs
```diff
@@ -551,3 +551,18 @@ rendy_without_metal_backend! {
         false
     }
 }
+
+
+rendy_with_gl_backend! {
```
Frizi: Why is this here?

mesh/src/mesh.rs
```diff
@@ -282,25 +282,47 @@ impl<'a> MeshBuilder<'a> {
...
+                            &mut buffer,
+                            0,
+                            &indices,
+                        )?;
+                    }
+                }
+                rendy_not_wasm32! {
+                    unsafe {
+                        // New buffer can't be touched by device yet.
+                        factory.upload_buffer(
```
Frizi: Why is the buffer upload done differently on wasm?

shader/src/shaderc.rs
```diff
@@ -70,8 +70,8 @@ where
                         .ok_or_else(|| failure::format_err!("Failed to init Shaderc"))?;
                     ops.set_target_env(shaderc::TargetEnv::Vulkan, vk_make_version!(1, 0, 0));
                     ops.set_source_language(self.lang);
-                    ops.set_generate_debug_info();
-                    ops.set_optimization_level(shaderc::OptimizationLevel::Performance);
+                    // ops.set_generate_debug_info();
+                    // ops.set_optimization_level(shaderc::OptimizationLevel::Performance);
```
Frizi: Why is optimization and debug info disabled? Debug info is required for reflection to work.

texture/src/texture.rs
```diff
@@ -265,34 +265,41 @@ impl<'a> TextureBuilder<'a> {
...
-                gfx_hal::image::Kind::D1(_, _) => (1, false),
-                gfx_hal::image::Kind::D2(w, h, _, _) => (mip_levels_from_dims(w, h), true),
-                gfx_hal::image::Kind::D3(_, _, _) => (1, false),
+                rendy_core::hal::image::Kind::D1(_, _) => (1, false),
+                rendy_core::hal::image::Kind::D2(w, h, _, _) => (mip_levels_from_dims(w, h), true),
+                rendy_core::hal::image::Kind::D3(_, _, _) => (1, false),
             },
         };
 
+        // gl backend doesn't support mips yet.
```
Frizi: Is this still true? If so, this should be marked as a TODO, ideally with tracking issue on hal side.

graph/src/node/mod.rs
omni-viral: For nodes to use.
Can be moved to core.

mesh/src/mesh.rs
omni-viral: Because copying as on desktop didn't work :)

shader/src/shaderc.rs
omni-viral: WebGL can't digest glsl generated from spirv compiled with those options.

handicraftsman:
What minimum OpenGL version is going to be supported? For example, i would like to support users with at least OpenGl 2.1 and OpenGl ES 2 (older is way too hard to implement from shader-based project).

mesh/src/mesh.rs
Frizi: It should. Is it just not implemented on webgl backend yet?

graph/src/node/mod.rs
Frizi: Yeah, moving that to core to be close to all the other backend specific macros and functions is a good idea.

shader/src/shaderc.rs
Frizi: Why is glsl output any different? The debug info should simply be ignored, maybe except as a suggestion for function and variable names. This sounds like a spirv-cross bug then and should be reported as such. Then, we should have a comment with link to the issue, and all of this should only be disabled on wasm.

fhaynes:
@omni-viral Please start breaking these PRs apart into much smaller ones. A PR with 3.8k lines changed and 136 files cannot be reviewed properly. There will be cases where this isn't possible, but the default, as we agreed some time ago, needs to be small, well-tested PRs.

Rather than close this one, let's pull out the crate tree modification into its own PR and get it merged.

We simply have to increase our code quality, and so far, I've seen very little of the procedural changes we talked about make it in. So we'll start with this one.

Let me know if you have any questions.

Frizi:
Splitting it off into separate PRs is definitely a good idea to manage this.

I think the best first candidate would be the modification of dependency tree (moving things to core, removing utils). This alone will be quite a big changeset, but very repetitive one. It should make reviewing the rest of stuff much easier. After that, the non-gl-related macro stuff would be a good candidate for a second PR.

omni-viral:
PR is subjet to division. This one stays open until digestible replacement prepared.

PR #191, opened 2019-08-26
changes for amethyst shadow mapping

Frizi:
This is a commit with changes I had to make while working on shadow mapping. The set includes:
- fix for broken serde feature after conversion to u32 spirv slices (`serde_bytes_u32`)
- removal of unnecessary assertion on spirv buffer length
- allow adding dynamic nodes to graph
- Allow creating `DescBuilder` with public `new`
- always pick first layer for bound attachments, comment explains why
- allow llvm inlining to better optimize texture conversion common cases

The rest is all pretty much just rustfmt. We probably need to add a CI step that verifies that, as we have quite a lot of noise due to not formatting the code immediately.

### DIFF
--- a/chain/src/collect.rs
+++ b/chain/src/collect.rs
@@ -1,6 +1,6 @@
 use std::cmp::max;
-use std::collections::HashMap;
 use std::collections::hash_map::RandomState;
+use std::collections::HashMap;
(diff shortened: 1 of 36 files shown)


Frizi:
bors r=omni-viral

PR #192, opened 2019-08-26
Clone transient resources iterator in render pass

### LINKED ISSUE #190: Two image attachments for SimpleGraphicsPipeline
Hi all!

Recently I've been trying to use Rendy in my little project (master thesis) and got some problems.

Once I try to use two images in my SimpleGraphicsPipeline implementation I've got panic during graph building in:

```
                if !link.access().exclusive() {
                    unimplemented!("This case is unimplemented");
                }
```
in chain/src/sync.rs: line 421

What I'm trying to achieve: render two images and then combine them in third pass.
```
    let pass3 = graph_builder.add_node(
        desc3.builder()
            .with_image(image1)
            .with_image(image2) // Once I add second image unimplemeted! is triggered
            .into_subpass()
            .with_dependency(pass1)
            .with_dependency(pass2)
            .with_color(final_output)
            .into_pass()
    );
```

Is there a way to work around this?
If anyone is willing to explain what the problem seems to be I can dig into Rendy and help implement missing functionality.

I couldn't find any example using more than one image used; if there is one I would love to see it.

Edit:

Actually this error was rather due to the fact that I wanted to use not written to image as an input image.

When actually using code I presented above I get different panic:

thread 'main' panicked at 'Transient image wasn't provided
in
node/render/pass.rs: 730.
```
                        let images: Vec<_> = group
                            .images()
                            .into_iter()
                            .map(|(id, _)| {
                                images
                                    .find(|i| i.id == id)
                                    .expect(format!("Transient image wasn't provided for {}", id))
                                    .clone()
                            })
                            .collect();
```

It looks like it can't find input image in map of all images. I'm investigating further.

Tsuguri:
Not cloned iterator was used multiple times causing possible errors
during images lookup.

The same applies to buffers.

Fixes #190

### DIFF
--- a/graph/src/node/render/pass.rs
+++ b/graph/src/node/render/pass.rs
@@ -724,14 +724,14 @@ where
                         assert_eq!(group.colors(), subpass_colors);
                         assert_eq!(group.depth(), subpass_depth);
 
-                        let mut buffers = buffers.iter();
-                        let mut images = images.iter();
+                        let buffers = buffers.iter();
+                        let images = images.iter();
 
                         let buffers: Vec<_> = group
                             .buffers()
                             .into_iter()
                             .map(|(id, _)| {
-                                buffers
+                                buffers.clone()
                                     .find(|b| b.id == id)
                                     .expect("Transient buffer wasn't provided")
                                     .clone()
@@ -741,7 +741,7 @@ where
                             .images()
                             .into_iter()
                             .map(|(id, _)| {
-                                images
+                                images.clone()
                                     .find(|i| i.id == id)
                                     .expect("Transient image wasn't provided")
                                     .clone()
--- a/graph/src/node/render/pass.rs
+++ b/graph/src/node/render/pass.rs
@@ -724,14 +724,11 @@ where
                         assert_eq!(group.colors(), subpass_colors);
                         assert_eq!(group.depth(), subpass_depth);
 
-                        let buffers = buffers.iter();
-                        let images = images.iter();
-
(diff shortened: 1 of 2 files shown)


posejdon256:
> Not cloned iterator was used multiple times causing possible errors
> during images lookup.
> 
> The same applies to buffers.
> 
> Fixes #190

Please add it to repository. It is crusial fix.

Tsuguri:
It seems that it automatically added second commit to the same pull request...

They're not really in the same topic, but still needed.

graph/src/node/render/pass.rs
```diff
@@ -724,14 +724,14 @@ where
...
-                        let mut images = images.iter();
+                        let buffers = buffers.iter();
+                        let images = images.iter();
 
                         let buffers: Vec<_> = group
                             .buffers()
                             .into_iter()
                             .map(|(id, _)| {
-                                buffers
+                                buffers.clone()
```
ishitatsuyuki: At this point, you could just use `buffers.iter()` and remove the temporary assignment above.

ishitatsuyuki:
If you're rushing with the fix, why not override your dependency locally for the meanwhile?

https://doc.rust-lang.org/cargo/reference/manifest.html#the-patch-section

graph/src/node/render/pass.rs
omni-viral: Just capture vectors directly. It will be more readable.

omni-viral: I second for this.

Issue #182, opened 2019-08-06
Image count not supported with vulkan backend on wayland.


krsnik02:
I am attempting to run the pong tutorial in the amethyst book and get the following error. I am using the vulkan backend in a wayland session in linux.
```
[WARN][rendy_wsi] Image count not supported. Supported: 4..4294967295, requested: 3
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ErrorMessage { msg: "Image count not supported." }
```
I believe I have tracked the source of the issue to line 443 in graph/src/node/render/pass.rs where a hardcoded 3 is passed as the image_count parameter to `rendy_factory::factory::Factory::create_target`. This should be changed to use the smaller of 3 and the minimum image count in the surface capabilities (which on my system is 4).

omni-viral:
`PresentNode` clamps default image count between supported values.
`RenderPassNode` should do the same.

malobre:
Is anyone working on this? I could take a look.

PR #193, opened 2019-08-29
Update pass.rs

malobre:
Fix https://github.com/amethyst/rendy/issues/182
Tested with some of the examples in rendy/examples.


omni-viral:
Nice! That's way better than hardcoded values.

bors r+

PR #195, opened 2019-08-30
Fix serde support

ishitatsuyuki:
Uses a hand-written serde implementation for `Vec<u32>`. Serializer is basically a copy of serde-bytes (plus the unsafe conversion), deserializer uses the endian-aware implementation from gfx-hal update.

### DIFF
--- a/shader/Cargo.toml
+++ b/shader/Cargo.toml
@@ -13,7 +13,7 @@ description = "Rendy's shader compilation tool"
 [features]
 shader-compiler = ["shaderc"]
 spirv-reflection = [ "spirv-reflect" ]
-serde-1 = ["serde", "serde_bytes"]
+serde-1 = ["serde", "gfx-hal/serde"]
 
 [dependencies]
 smallvec = "0.6"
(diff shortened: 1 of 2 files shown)


ishitatsuyuki:
ping

Frizi:
I already have that done in https://github.com/amethyst/rendy/pull/191, so this is unfortunately a duplicate :(

Frizi:
Although your deserialize implementation is in fact more correct, so I'll copy it over instead. Whatever is merged first I guess.

Frizi:
I have in fact additionaly used serde_bytes for it's performance. Please look at #191 instead.

ishitatsuyuki:
Well... So my implementation is actually a clone of serde_bytes, and serde_bytes doesn't do anything that relates to performance. [It's just a one-liner.](https://github.com/serde-rs/bytes/blob/master/src/ser.rs)

Frizi:
Oh well, good catch :D So all it does is just forces the usage of serialize_bytes instead of going through the slice as collection. I'll inline that then and get rid of the dependency. Thanks!

ishitatsuyuki:
I'm fine with going with that other PR, although I believe git will go without merge conflicts if the patch is identical. Since this is a blocker for Rendy upgrade in Amethyst, do you think it's better to merge this beforehand?

ishitatsuyuki:
Closing as #191 is merged.

PR #196, opened 2019-08-30
commands: Add dynamic state related wrapper functions

ishitatsuyuki:
I left out non-render-pass image related functions, like ClearColorImage or BlitImage. Maybe they can be added if there's a demand?

### DIFF
--- a/command/src/buffer/encoder.rs
+++ b/command/src/buffer/encoder.rs
@@ -258,6 +258,20 @@ where
         );
     }
 
+    /// Set viewports
+    ///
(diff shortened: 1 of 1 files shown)


ishitatsuyuki:
I will fix imports and errors later.

ishitatsuyuki:
As I described in PR body, can someone tell me what are the docs supposed to be? They doesn't match the ones from gfx-hal RawCommandBuffer, and copying the respective part from Vulkan spec seems to be too much work.

omni-viral:
There should be link to respective vulkan spec page at least

PR #197, opened 2019-08-30
Fixes #58 + minor fixes

### LINKED ISSUE #58: Request: `T: 'static + Copy` bound on Factory's upload_* functions
I just ran into an issue where I was accidentally uploading some references to the GPU - passing in `&[some_ref]` to `upload_visible_buffer`. The `'static` bound would have prevented that error.

malobre:
Fixes #58 + minor changes:
* The `source_shaders` example had incorrect logging filter
* The `quads` example used `println!` instead of logging-macros

### DIFF
--- a/factory/src/factory.rs
+++ b/factory/src/factory.rs
@@ -398,16 +398,24 @@ where
         &self,
         buffer: &mut Buffer<B>,
         offset: u64,
-        content: &[T],
-    ) -> Result<(), MapError> {
+        content: &T,
+    ) -> Result<(), MapError>
+    where
+        T: ?Sized + 'static,
+    {
+        let content_size = std::mem::size_of_val(content);
+
         let content = std::slice::from_raw_parts(
-            content.as_ptr() as *const u8,
-            content.len() * std::mem::size_of::<T>(),
+            {
+                let content_ptr: *const T = content;
+                content_ptr as *const u8
+            },
+            content_size,
         );
 
-        let mut mapped = buffer.map(&self.device, offset..offset + content.len() as u64)?;
+        let mut mapped = buffer.map(&self.device, offset..offset + content_size as u64)?;
         mapped
-            .write(&self.device, 0..content.len() as u64)?
+            .write(&self.device, 0..content_size as u64)?
             .write(content);
         Ok(())
     }
@@ -437,13 +445,17 @@ where
         &self,
         buffer: &Buffer<B>,
         offset: u64,
-        content: &[T],
+        content: &T,
         last: Option<BufferState>,
         next: BufferState,
-    ) -> Result<(), UploadError> {
+    ) -> Result<(), UploadError>
+    where
+        T: ?Sized + 'static,
+    {
         assert!(buffer.info().usage.contains(buffer::Usage::TRANSFER_DST));
 
-        let content_size = content.len() as u64 * std::mem::size_of::<T>() as u64;
+        let content_size = std::mem::size_of_val(content) as u64;
+
         let mut staging = self
             .create_buffer(
                 BufferInfo {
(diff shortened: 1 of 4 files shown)


omni-viral:
While you're at it.
It would be more flexible without any overhead to use `content: &T` where `T: ?Sized + 'static`. This way we allow both singular values, slices and custom DSTs.
`Copy` is not desirable here as it can't be implemented for types with big arrays.

malobre:
Will do.

omni-viral:
@malobre You don't need to worry about that.
Just put `?Sized` bound to allow DST and remove `Copy` bound to allow uncopiable types (like `[u8; 33]`).
And finally replace `&[T]` with `&T`. As with `T: ?Sized` it will accept references to slices.

malobre:
But how would you go about getting content size in, for example, `upload_visible_buffer(...)` ? `T` becomes potentially unsized, and, to my knowledge, there is no runtime function to get a DST size.

omni-viral:
[This function](https://doc.rust-lang.org/stable/std/mem/fn.size_of_val.html) does exactly that.

malobre:
Don't know how I missed that. 🤦‍♂️
I modified every `upload_...` functions but I've got a lifetime problem in [mesh.rs](https://github.com/malobre/rendy/blob/5689a516d98f2ef18c8083aa2ed5341dbe64723a/mesh/src/mesh.rs#L298):
```
error[E0495]: cannot infer an appropriate lifetime for lifetime parameter `'a` due to conflicting requirements
   --> mesh/src/mesh.rs:281:18
    |
281 |             Some(RawIndices {
    |                  ^^^^^^^^^^
    |
note: first, the lifetime cannot outlive the lifetime 'a as defined on the impl at 113:6...
   --> mesh/src/mesh.rs:113:6
    |
113 | impl<'a> MeshBuilder<'a> {
    |      ^^
    = note: ...so that the types are compatible:
            expected mesh::RawIndices<'a>
               found mesh::RawIndices<'_>
    = note: but, the lifetime must be valid for the static lifetime...
note: ...so that the type `&std::borrow::Cow<'_, [u8]>` will meet its required lifetime bounds
   --> mesh/src/mesh.rs:295:29
    |
295 |                     factory.upload_buffer(
    |                             ^^^^^^^^^^^^^

```

omni-viral:
You got this error after changing `&[T]` to `&T`, right?

If argument is `&'inferred [T]` then `&'inferred Cow<'a, [U]>` coerces to `&'inferred [U]` and `T = U`.
If argument is `&'inferred T` then `&'inferred Cow<'a, [U]>` not get coerced and `T = Cow<'a, [U]>`, so `T: 'static` bound is not satisfied.

This means that `[T] -> T` change is breaking and where implicit coercion was used explicit `*` would be required. i.e. `&indices -> &*indices,`

malobre:
Thanks for the explanation.
I had to add an extra `*` since `indices` is already a reference.
It seems to be working, I'll resolve conflicts soon.

malobre:
I am unable to make the `quads` example work (see previous comment).
Every other example are running fine.

malobre:
The Travis CI build failed because it hadn't enough time to build on OSX.

factory/src/factory.rs
```diff
@@ -393,20 +393,25 @@ where
...
     pub unsafe fn upload_visible_buffer<T>(
         &self,
         buffer: &mut Buffer<B>,
         offset: u64,
-        content: &[T],
-    ) -> Result<(), MapError> {
+        content: &T,
+    ) -> Result<(), MapError>
+    where
+        T: ?Sized + 'static,
```
omni-viral: There is no such thing as `!Drop` and we can't use `Copy` (it implies `!Drop`) due to large arrays being `!Copy`.
But at least we can assert that `std::mem::needs_drop::<T>() == false;`
We could even use some form of static assertions.

malobre: `std::mem::needs_drop::<T>()` requires `T: Sized`

factory/src/factory.rs
```diff
@@ -393,20 +393,25 @@ where
...
+    {
+        let content_size = std::mem::size_of_val(content);
         let content = std::slice::from_raw_parts(
-            content.as_ptr() as *const u8,
-            content.len() * std::mem::size_of::<T>(),
+            content as *const T as *const u8,
```
malobre: This trigger a trivial cast warning, I couldn't figure out another way so I allowed trivial_casts for this function.

factory/src/factory.rs
```diff
@@ -436,13 +441,17 @@ where
...
         next: BufferState,
-    ) -> Result<(), UploadError> {
+    ) -> Result<(), UploadError>
+    where
+        T: ?Sized + 'static,
+    {
         assert!(buffer.info().usage.contains(buffer::Usage::TRANSFER_DST));
 
-        let content_size = content.len() as u64 * std::mem::size_of::<T>() as u64;
+        let content_size = std::mem::size_of_val(content) as _;
```
malobre: Should this be more explicit ?
```suggestion
        let content_size = std::mem::size_of_val(content) as u64;
```

factory/src/factory.rs
omni-viral: Did you try `content as *const u8`

malobre: Yes, doesn't work.
```
error[E0606]: casting `&T` as `*const u8` is invalid
   --> factory/src/factory.rs:409:50
    |
409 |         let content = std::slice::from_raw_parts(content as *const u8, content_size);
    |                                                  ^^^^^^^^^^^^^^^^^^^^
```

omni-viral: I know this is silly, but could you use this
```
std::slice::from_raw_parts({let c: *const _ = content; c as *const u8}, content_size)
```

because this is public function and we should keep our dirty loundry in implementation details :)

malobre: No problem, I'll commit this tonight.

omni-viral:
Could you please rebase? :)

malobre:
@omni-viral Done

PR #198, opened 2019-09-01
Update to master gfx-rs

### LINKED ISSUE #194: Update Rendy to use gfx-rs from git
In wgpu, we want to test out some of the features that landed after gfx-0.3. Namely, the new swapchain model. Would it be possible to get Rendy `master` to depend on gfx-rs from a recent git revision (instead of crates.io)?

kvark:
Fixes  #194
Relying on `failure` is a real pain :/

### DIFF
--- a/chain/Cargo.toml
+++ b/chain/Cargo.toml
@@ -14,6 +14,6 @@ description = "Rendy's node synchronization tool"
 profiler = ["thread_profiler/thread_profiler"]
 
 [dependencies]
-gfx-hal = "0.3"
+gfx-hal = { git = "https://github.com/gfx-rs/gfx", rev = "a084a1f2fec6c9ed928c53e2dc8c1761782b9019" }
 log = "0.4"
 thread_profiler = "0.3"
--- a/command/Cargo.toml
+++ b/command/Cargo.toml
@@ -15,7 +15,7 @@ no-slow-safety-checks = ["rendy-util/no-slow-safety-checks"]
 profiler = ["thread_profiler/thread_profiler"]
 
 [dependencies]
-gfx-hal = "0.3"
+gfx-hal = { git = "https://github.com/gfx-rs/gfx", rev = "a084a1f2fec6c9ed928c53e2dc8c1761782b9019" }
 derivative = "1.0"
 failure = "0.1"
 relevant = { version = "0.4.0", features = ["log", "backtrace"] }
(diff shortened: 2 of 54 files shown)


kvark:
@omni-viral any tips/preference on how to get this finished?

omni-viral:
Looks nice. Let's make CI happy :)

kvark:
@omni-viral how about now?

kvark:
Ok, now it should properly build on CI.
@omni-viral I'm not super happy about the part `Some(graph)`. Had to do it because `EventLoop::run()` is not returning any more and it's closure takes ownership of things. This could probably be written much nicer, e.g. by not moving things into that closure.

kvark:
Another annoying thing is that the meshes example panics on exit (in Factory::drop). I see that the original code force the drop order of Families versus Factory, but trying to do the same with the new code doesn't seem to help.

kvark:
I certainly didn't intend to invest into the whole winit update thing :/
@omni-viral please have a look

kvark:
bors retry

kvark:
timeouts :/
bors retry

kvark:
@omni-viral can we has higher CI timeouts?..

kvark:
bors retry

kvark:
I wish we could say something like "bors retry timeout=50", supposing that "timeout" is just a user-defined variable in bors with some default value, which it can propagate (somehow) to the actual CI services.
cc @notriddle

notriddle:
1. You can prevent Travis from timing out by using [`travis_wait`](https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received)

2. You can prevent bors from timing out by using [`timeout-sec`](https://bors.tech/documentation/#configuration-borstoml)

kvark:
bors retry

omni-viral:
@kvark IIRC we already set it to 30

Frizi:
bors retry

kvark:
Bors retry 
...

notriddle:
Hey, @kvark 

Did you notice that [Travis CI's status page](https://www.traviscistatus.com/) reports degraded service? You might want to wait until they fix that.

kvark:
bors retry

kvark:
bors retry

kvark:
It doesn't look like the status means much for this case. What matters more is the backlog size, and the fact current timeouts are abysmally small.

kvark:
bors retry

kvark:
bors retry

omni-viral:
@kvark maybe we can just merge it? :)

kvark:
Yes please

PR #200, opened 2019-09-04
Fix texture corruption due to buffer mis-align

cloudhead:
This patch ensures that even when the dynamic allocator
fulfills an allocation request by returning multiple small
blocks, the memory is correctly aligned.

This wasn't the case previously, as alignment was only
guaranteed to match the block size, so when fulfilling
a large allocation with multiple smaller blocks, the memory
would be aligned to the smaller block size.

The validation triggered by the Vulkan layer before this fix
would look like this:

    VALIDATION [VUID-vkBindImageMemory-memoryOffset-01048 (0)]:
    vkBindImageMemory(): memoryOffset is 0x43300 but must be an integer
                         multiple of the VkMemoryRequirements::alignment
                         value 0x1000, returned from a call to
                         vkGetImageMemoryRequirements with image. The
                         Vulkan spec states: memoryOffset must be an
                         integer multiple of the alignment member of the
                         VkMemoryRequirements structure returned from a
                         call to vkGetImageMemoryRequirements [...]

### DIFF
--- a/memory/src/allocator/dynamic.rs
+++ b/memory/src/allocator/dynamic.rs
@@ -307,14 +307,14 @@ where
             .next_back()
         {
             // Allocate block for the chunk.
-            let (block, allocated) = self.alloc_from_entry(device, chunk_size, 1)?;
+            let (block, allocated) = self.alloc_from_entry(device, chunk_size, 1, block_size)?;
             Ok((Chunk::from_block(block_size, block), allocated))
         } else {
             let total_blocks = self.sizes[&block_size].total_blocks;
             let chunk_size =
                 (max_chunk_size.min(min_chunk_size.max(total_blocks * block_size)) / 2 + 1)
                     .next_power_of_two();
-            let (block, allocated) = self.alloc_block(device, chunk_size)?;
+            let (block, allocated) = self.alloc_block(device, chunk_size, block_size)?;
             Ok((Chunk::from_block(block_size, block), allocated))
         }
     }
@@ -325,6 +325,7 @@ where
         chunk_index: u32,
         block_size: u64,
         count: u32,
+        align: u64,
     ) -> Option<DynamicBlock<B>> {
         log::trace!(
             "Allocate {} consecutive blocks of size {} from chunk {}",
@@ -334,8 +335,9 @@ where
         );
 
         let ref mut chunk = chunks[chunk_index as usize];
-        let block_index = chunk.acquire_blocks(count)?;
+        let block_index = chunk.acquire_blocks(count, block_size, align)?;
         let block_range = chunk.blocks_range(block_size, block_index, count);
+
         debug_assert_eq!((block_range.end - block_range.start) % count as u64, 0);
 
         Some(DynamicBlock {
@@ -358,6 +360,7 @@ where
         device: &B::Device,
         block_size: u64,
         count: u32,
+        align: u64,
     ) -> Result<(DynamicBlock<B>, u64), gfx_hal::device::AllocationError> {
         log::trace!(
             "Allocate {} consecutive blocks for size {} from the entry",
@@ -369,9 +372,13 @@ where
         let size_entry = self.sizes.entry(block_size).or_default();
 
         for chunk_index in (&size_entry.ready_chunks).iter() {
-            if let Some(block) =
-                Self::alloc_from_chunk(&mut size_entry.chunks, chunk_index, block_size, count)
-            {
+            if let Some(block) = Self::alloc_from_chunk(
+                &mut size_entry.chunks,
+                chunk_index,
+                block_size,
+                count,
+                align,
+            ) {
                 return Ok((block, 0));
             }
         }
@@ -385,8 +392,14 @@ where
         let size_entry = self.sizes.entry(block_size).or_default();
         let chunk_index = size_entry.chunks.insert(chunk) as u32;
 
-        let block = Self::alloc_from_chunk(&mut size_entry.chunks, chunk_index, block_size, count)
-            .expect("New chunk should yield blocks");
+        let block = Self::alloc_from_chunk(
+            &mut size_entry.chunks,
+            chunk_index,
+            block_size,
+            count,
+            align,
+        )
+        .expect("New chunk should yield blocks");
 
         if !size_entry.chunks[chunk_index as usize].is_exhausted() {
             size_entry.ready_chunks.add(chunk_index);
@@ -400,6 +413,7 @@ where
         &mut self,
         device: &B::Device,
         block_size: u64,
+        align: u64,
     ) -> Result<(DynamicBlock<B>, u64), gfx_hal::device::AllocationError> {
         log::trace!("Allocate block of size {}", block_size);
 
@@ -415,7 +429,12 @@ where
                 .range(block_size / 4..block_size * overhead)
                 .next()
             {
-                return self.alloc_from_entry(device, size, ((block_size - 1) / size + 1) as u32);
+                return self.alloc_from_entry(
+                    device,
+                    size,
+                    ((block_size - 1) / size + 1) as u32,
+                    align,
+                );
             }
         }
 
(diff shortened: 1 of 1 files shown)


cloudhead:
Tentative fix for https://github.com/gfx-rs/wgpu/issues/323 and https://github.com/cloudhead/rx/issues/14

cloudhead:
I have only very basic knowledge of how this library works, so maybe my fix is naive :man_shrugging:

memory/src/allocator/dynamic.rs
```diff
@@ -280,6 +280,7 @@ where
         device: &B::Device,
         block_size: u64,
         total_blocks: u64,
+        align: u64,
```
omni-viral: Allocated chunk will be used for other allocations, while this value carries alignment requirement of one particular request.
And it will be aligned to its size (not block size, but whole chunks size) anyway.
Its impossible to serve allocation of size N using blocks from chunk smaller than N.
Plus, allocation from smaller blocks should never try to allocate new chunk.

TL;DR; this value shouldn't ever happen to be greater than chunk size, thus never increase allocation alignment, thus not needed here.

memory/src/allocator/dynamic.rs
```diff
@@ -336,6 +338,9 @@ where
         let ref mut chunk = chunks[chunk_index as usize];
         let block_index = chunk.acquire_blocks(count)?;
         let block_range = chunk.blocks_range(block_size, block_index, count);
+        if block_range.start % align != 0 {
```
omni-viral: This check will result in allocation failure when this chunk could provide blocks with required alignment.
For example:
Chunk has size `128` and blocks are of size `4`. Let's say chunk starts at offset `0`.
We request `4` blocks with alignment `16`.
If first blocks in chunk was allocated then we'd get `block_range = 4..20`. It doesn't have required alignment.
But we could allocate range `16..32` from the chunk instead.

cloudhead: This makes sense, yes! I've tried to fix this here f5f7a2e - let me know if that is what you meant. The only question I have is whether the `block_size` is correct here, since it's passed as a parameter, and I'm not sure how to verify if/how it matches the chunk's block_size.

memory/src/allocator/dynamic.rs
cloudhead: Ah I see, will fix :+1:

cloudhead: Fixed here: f72edb9 

I used `Option`, but passing an `align` of `1` would also work.

memory/src/allocator/dynamic.rs
```diff
@@ -611,20 +630,21 @@ where
...
             blocks &= self.blocks >> i;
         }
         let index = blocks.trailing_zeros();
-        if index == MAX_BLOCKS_PER_CHUNK {
-            None
-        } else {
-            let mask = ((1 << count) - 1) << index;
-            self.blocks &= !mask;
-            Some(index)
+        for i in index..MAX_BLOCKS_PER_CHUNK {
```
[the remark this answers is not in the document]
cloudhead: I'm realizing now that the code is wrong, but that the original code also doesn't make sense to me: it finds an index into the bit-array where there is a `1`, ie. a free slot, then creates a mask for `count` bits, but *assumes* that there are `count` * `1` bits at `index`, which isn't verified anywhere? Am I missing something?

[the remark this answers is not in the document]
cloudhead: Ah nevermind, I get it - that's what the first loop does :+1:

memory/src/allocator/dynamic.rs
```diff
@@ -307,14 +307,14 @@ where
             .next_back()
         {
             // Allocate block for the chunk.
-            let (block, allocated) = self.alloc_from_entry(device, chunk_size, 1)?;
+            let (block, allocated) = self.alloc_from_entry(device, chunk_size, 1, None)?;
```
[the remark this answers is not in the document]
cloudhead: Done here 47e0b35

memory/src/allocator/dynamic.rs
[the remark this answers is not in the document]
cloudhead: Getting this to work is much more complicated than I expected. I think the new code here: 6c3313a is correct, but it's more complicated. I'm not sure how to simplify it, or if it's possible to figure out the right index without trying each index, eg. if `block_size = 432411` and `align = 16`, I don't know of a simple formula to solve for `n` in `(432411 * n) % 16 = 0`. If `block_size < align`, then it's fairly easy though, but that's a special case.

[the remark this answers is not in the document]
cloudhead: Furthermore, there is a case where for eg. you have `blocks = 00000111100`, and `count = 4`, `index = 2`, which would work fine if it was all aligned. But now imagine `index = 3` is the one that's correctly aligned, but there aren't enough blocks to fulfill that. That's why I'm checking each bit in the range `3..3+count` here. Again, I'm sure there's a faster way, but I want to make sure this is even correct.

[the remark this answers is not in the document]
cloudhead: Ok, I cleaned things up a bit - here's the current state: https://github.com/cloudhead/rendy/blob/master/memory/src/allocator/dynamic.rs#L633-L660

I'm not sure what you meant by making it in one loop - did you mean combining the loops together?

memory/src/allocator/dynamic.rs
```diff
@@ -611,20 +630,33 @@ where
...
-            self.blocks &= !mask;
-            Some(index)
+
+        // We may not be aligned at `index`, but may find alignment
+        // at subsequent indices.
+        for i in index..MAX_BLOCKS_PER_CHUNK {
+            let mask: u64 = ((1 << count) - 1) << i;
+
+            // Exit if we don't have enough free blocks at the current offset.
+            if (self.blocks & mask).count_ones() < count {
```
omni-viral: This check can be simplified to `self.blocks & mask == mask`.
Or alternatively `blocks & (1 << i) == (1 << i)`. So instead of looping over every index we can just iterate over 1s in `blocks`.

memory/src/allocator/dynamic.rs
```diff
@@ -611,20 +630,33 @@ where
...
+                return None;
```
omni-viral: If check fails we should look further instead of returning.
For example if `self.blocks` is `b000110110` and we need `2` blocks with `align = 2 * block_size` then first pair doesn't fit, but this check will fail at next indicex (and at one after it'd fail too) and another free pair that statisfies requirements won't be examined.

memory/src/allocator/dynamic.rs
```diff
@@ -611,20 +630,33 @@ where
...
-        if index == MAX_BLOCKS_PER_CHUNK {
-            None
-        } else {
-            let mask = ((1 << count) - 1) << index;
-            self.blocks &= !mask;
-            Some(index)
+
+        // We may not be aligned at `index`, but may find alignment
+        // at subsequent indices.
+        for i in index..MAX_BLOCKS_PER_CHUNK {
```
omni-viral: No reason to look past `MAX_BLOCKS_PER_CHUNK - count`.

memory/src/allocator/dynamic.rs
cloudhead: Ah yes, good point. `count_ones` is cheap though, as it uses the `popcount` instruction.

memory/src/allocator/dynamic.rs
```diff
@@ -611,20 +630,29 @@ where
...
-        if index == MAX_BLOCKS_PER_CHUNK {
-            None
-        } else {
-            let mask = ((1 << count) - 1) << index;
-            self.blocks &= !mask;
-            Some(index)
+
+        // We may not be aligned at `index`, but may find alignment
+        // at subsequent indices.
+        for i in index..MAX_BLOCKS_PER_CHUNK - count {
```
omni-viral: Iterating over indices of set bits would be faster than iterating over all and checking mask.
And checking alignment doesn't require division (% operator is compiled to division instruiction).

So I think this loop can be rewritten as

```rust
while blocks != 0 {
  let index = blocks.trailing_zeros();
  blocks &= !(1 << index);
  if (index as u64 * block_size) & (align - 1) == 0 {
    return Some(index);  
  }
}
```

cloudhead: Done. I had to also mask `self.blocks` - maybe there's a way to reuse the other mask?

PR #206, opened 2019-09-18
Use expect instead of unwrap for vertex shader

bschwind:
While creating a minimal rendy example, I first started off with a pipeline with no shaders attached. I got a `thread 'main' panicked at 'called `Option::unwrap()` on a `None` value'` error from `shader/src/lib.rs:121`

Though probably an uncommon usage of rendy, it's a bit nicer to provide reasoning for why the program panicked.

### DIFF
--- a/shader/src/lib.rs
+++ b/shader/src/lib.rs
@@ -121,7 +121,7 @@ impl<B: Backend> ShaderSet<B> {
             vertex: self
                 .shaders
                 .get(&ShaderStageFlags::VERTEX)
-                .unwrap()
+                .expect("ShaderSet doesn't contain vertex shader")
                 .get_entry_point()?
                 .unwrap(),
             fragment: match self.shaders.get(&ShaderStageFlags::FRAGMENT) {


shader/src/lib.rs
```diff
@@ -121,7 +121,7 @@ impl<B: Backend> ShaderSet<B> {
             vertex: self
                 .shaders
                 .get(&ShaderStageFlags::VERTEX)
-                .unwrap()
+                .expect("No vertex shader was found in the GraphicsShaderSet")
```
omni-viral: Message is slightly misleading.
We build `GraphicsShaderSet`. Vertex shader wasn't found in `ShaderSet`.
I'm not English expert, but it seems that message also implies that there can be multiple vertex shaders which is not the case in shader set.

bschwind: Ah you're right, I'll replace `GraphicsShaderSet` with just `ShaderSet`

As for your other concern, I don't think there's a strong implication of multiple vertex shaders being allowed, and the "one shader per type" rule is also implied by the names and types of the struct.

The important message is: "We tried to build a GraphicsShaderSet but the ShaderSet did not contain a vertex shader, which is required". Would you prefer that?

omni-viral: I'd go with simple `"ShaderSet doesn't contain vertex shader"`

bschwind:
@omni-viral updated with your suggestion, sorry for the back-and-forth.

PR #209, opened 2019-09-21
Switch to jenkins CI

### LINKED ISSUE #208: move continuous integration stack to amethyst jenkins
Currently the CI checks are done on travis, which unfortunately times out very often on this workload. It needs to be moved to amethyst's jenkins. We already have jenkinsfile ready in this repositry, so I believe it's mostly about setting up the right github integration.

The current jenkins pipeline doesn't test on macOS and iOS devices. We need to figure out how critical it is and either enable macos testing or check if we can get away with not having it right now.

Frizi:
Fixes #208

iOS builds are still happening on travis, but only on master branch as a sanity check (does not block merge).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: rust
 compiler: clang
+cache: cargo
 
 env:
   global: RUSTFLAGS="-D warnings" RUST_LOG=info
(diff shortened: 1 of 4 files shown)


Frizi:
> Is there an alternative to an iOS device than travis?
An iOS device on jenkins :P

Frizi:
bors r=fhaynes

Frizi:
bors abort

Frizi:
bors r=fhaynes

Issue #203, opened 2019-09-10
Optimize build times


kvark:
Removing `failure` and `derivative` should be a solid step in this direction. See https://github.com/gfx-rs/gfx/pull/2970 and also https://github.com/amethyst/rendy/pull/198

Veykril:
As already said in #210, removing `derivative` does not seem to have any major effect in build times. I've just tested out how `failure` affects the build times by removing it and replacing the `failure::Errors` with basic appropriate structs/enums. Given how sparse it is actually used it didnt seem to have any major effect on build times either, resulting in `24.29s` build time for just the rendy crates and `1m 04s` for a full build, compared to the previous `24.53s` and `1m 08s`. The 4 second cut off is most likely due to `failure` not being a dependency in the whole project anymore thanks to the removal. (All the timings were done with possible background noise though, which is why i had one rendy crates-only build with just 20s).

Nevertheless I'd say to remove `failure` given that it seems to have been just used as a convenient error wrapping for strings. One or two rendy crates even have it as a left over dependency without even using it in the source like `rendy-frame` for example.

So if the consensus is to remove `failure` I'll be happy to clean up my test changes and make a PR for it.

kvark:
#198 has largely removed `failure`, it's just a few places left before it can be expelled from the Cargo toml. That, I think, explains why you aren't seeing much gains from finishing this work.
Overall, it would be great to see your PR doing this. Derivative can stay, for now :)

malobre:
Compile time report from [cargo-bloat](https://crates.io/crates/cargo-bloat):
```
$ cargo bloat --release --time -j 1
Compiling ...

  Time Crate
53.56s serde_derive
36.94s winit
36.69s syn
34.63s syn
32.71s wayland_protocols
31.66s derivative
27.23s smithay_client_toolkit
25.83s syn
18.27s synstructure
17.74s wayland_client
17.18s wayland_sys
15.61s gfx_hal
13.25s serde
12.90s x11_dl
12.37s wayland_scanner
12.17s nix
12.12s rendy_graph
11.86s cc
10.43s rendy_command
 9.96s rustc_serialize
 9.50s rendy_shader
 8.53s proc_macro2
 8.52s proc_macro2
 8.40s proc_macro2
 8.28s xml
 8.03s serde_json
 7.99s rendy_factory
 7.91s rendy_memory
 7.57s andrew
 7.43s rendy_util
 7.18s failure_derive
 7.08s spirv_reflect
[...]

PR #210, opened 2019-09-22
Remove derivative from rendy

Veykril:
Removes the derivative dependency from crates that are using it, as mentioned in #203, replacing it with manual implementations of the corresponding traits.

### DIFF
--- a/command/Cargo.toml
+++ b/command/Cargo.toml
@@ -16,8 +16,6 @@ profiler = ["thread_profiler/thread_profiler"]
 
 [dependencies]
 gfx-hal = { git = "https://github.com/gfx-rs/gfx", rev = "3d5db15661127c8cad8d85522a68ec36c82f6e69" }
-derivative = "1.0"
-failure = "0.1"
 relevant = { version = "0.4.0", features = ["log", "backtrace"] }
 smallvec = "0.6"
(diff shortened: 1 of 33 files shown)


Frizi:
I like the idea, but before we commit to writing impls manually i'd like to see actual numbers. Can you measure the build times from before and after this change?

Veykril:
Sure, and as it turns out the times are pretty much the exact same as it seems, regarding just building the rendy crates without the dependencies I got `24.53s` for 9e7cb219dda4426dab9abebdf32cd2353b28862d and `24.29s` for my branch, and with all dependencies its `1m 08s` for both, though for my branch derivative is still being build as dependency due to winit using it. So it seems derivative itself doesnt do much.

Frizi:
In that case i would prefer to get rid of `failure` first. Derivative macros are really useful for maintenance reasons, so if there is no real difference in build times I'd like to keep them.

Veykril:
Ye figured, definitely the smarter choice to stick with derivative then for the time being. Will try to see what removing failure will do and if it improves the times or not, depending on that I'll do a PR or comment in issue #203 with my findings.

Frizi:
Okay, thank you for doing that @Veykril :)

PR #211, opened 2019-09-23
remove failure

Veykril:
This PR removes the last bits of the `failure` crate. There are a few things I wasn't sure about which I'm just gonna mention here(they all address `rendy-shader`). 

The `ShaderError` enum variants depend on features which can change the amount of variants of it, due to that I added an `__Unexhaustive` variant so toggling features on won't suddenly break matching code, whether this is the correct way to do this or not is something I don't know. Also regarding this enum's name, it clashes with the `ShaderError` of `gfx_hal`, which both are used in the `Shader` trait now, so maybe gfx_hals shader error should become a variant of this one so that the Shader trait returns the same error type for all of its functions preventing potential confusion.
The last point is that the errors for the shader crate are rather nested now which is probably not needed, but I wasn't certain what other way might be the preferred one here.

Note: it's also a shame that the `spirv-reflect` errors arent being returned as `'static` string slices, which they internally are, which forces the strings to be cloned so that they can be propagated. I already made a PR propagate the `'static` lifetime for them there(https://github.com/gwihlidal/spirv-reflect-rs/pull/9).

### DIFF
--- a/mesh/Cargo.toml
+++ b/mesh/Cargo.toml
@@ -25,7 +25,6 @@ rendy-util = { version = "0.4.0", path = "../util" }
 
 gfx-hal = { git = "https://github.com/gfx-rs/gfx", rev = "3d5db15661127c8cad8d85522a68ec36c82f6e69" }
 
-failure = "0.1"
 serde = { version = "1.0", optional = true, features = ["derive"] }
 wavefront_obj = { version = "6.0", optional = true }
 smallvec = { version = "0.6" }
--- a/mesh/src/format/obj.rs
+++ b/mesh/src/format/obj.rs
@@ -7,23 +7,19 @@ use {
 };
 
 /// Load mesh data from obj.
-pub fn load_from_obj(
-    bytes: &[u8],
-) -> Result<Vec<(MeshBuilder<'static>, Option<String>)>, failure::Error> {
-    let string = std::str::from_utf8(bytes)?;
-    let set = obj::parse(string).map_err(|e| {
-        failure::format_err!(
+pub fn load_from_obj(bytes: &[u8]) -> Result<Vec<(MeshBuilder<'static>, Option<String>)>, String> {
+    let string = std::str::from_utf8(bytes).map_err(|e| e.to_string())?;
+    obj::parse(string).and_then(load_from_data).map_err(|e| {
+        format!(
             "Error during parsing obj-file at line '{}': {}",
-            e.line_number,
-            e.message
+            e.line_number, e.message
         )
-    })?;
-    load_from_data(set)
+    })
 }
 
 fn load_from_data(
     obj_set: obj::ObjSet,
-) -> Result<Vec<(MeshBuilder<'static>, Option<String>)>, failure::Error> {
+) -> Result<Vec<(MeshBuilder<'static>, Option<String>)>, wavefront_obj::ParseError> {
     // Takes a list of objects that contain geometries that contain shapes that contain
     // vertex/texture/normal indices into the main list of vertices, and converts to
     // MeshBuilders with Position, Normal, TexCoord.
--- a/mesh/src/format/obj.rs
+++ b/mesh/src/format/obj.rs
@@ -6,15 +6,37 @@ use {
     wavefront_obj::obj,
 };
 
+/// Object loading error.Option
+#[derive(Debug)]
+pub enum ObjError {
+    /// The passed bytes were improper UTF-8 data.
+    Utf8(std::str::Utf8Error),
+    /// Parsing of the obj failed.
+    Parse(wavefront_obj::ParseError),
+}
+
+impl std::error::Error for ObjError {}
+impl std::fmt::Display for ObjError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            ObjError::Utf8(e) => write!(f, "{}", e),
+            ObjError::Parse(e) => write!(
+                f,
+                "Error parsing object file at line {}: {}",
+                e.line_number, e.message
+            ),
+        }
+    }
+}
+
 /// Load mesh data from obj.
-pub fn load_from_obj(bytes: &[u8]) -> Result<Vec<(MeshBuilder<'static>, Option<String>)>, String> {
-    let string = std::str::from_utf8(bytes).map_err(|e| e.to_string())?;
-    obj::parse(string).and_then(load_from_data).map_err(|e| {
-        format!(
-            "Error during parsing obj-file at line '{}': {}",
-            e.line_number, e.message
-        )
-    })
+pub fn load_from_obj(
+    bytes: &[u8],
+) -> Result<Vec<(MeshBuilder<'static>, Option<String>)>, ObjError> {
+    let string = std::str::from_utf8(bytes).map_err(ObjError::Utf8)?;
+    obj::parse(string)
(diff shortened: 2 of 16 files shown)


mesh/src/format/obj.rs
```diff
@@ -7,23 +7,19 @@ use {
 };
 
 /// Load mesh data from obj.
-pub fn load_from_obj(
-    bytes: &[u8],
-) -> Result<Vec<(MeshBuilder<'static>, Option<String>)>, failure::Error> {
-    let string = std::str::from_utf8(bytes)?;
-    let set = obj::parse(string).map_err(|e| {
-        failure::format_err!(
+pub fn load_from_obj(bytes: &[u8]) -> Result<Vec<(MeshBuilder<'static>, Option<String>)>, String> {
```
omni-viral: Could you kindly make an enum for those two error types instead of converting them into string?

Veykril: Ah, missed that one my bad!

mesh/Cargo.toml
```diff
@@ -15,7 +15,7 @@ readme = "README.md"
 obj = ["wavefront_obj"]
 serde-1 = ["serde", "serde_bytes", "gfx-hal/serde", "smallvec/serde", "rendy-factory/serde-1"]
 no-slow-safety-checks = ["rendy-util/no-slow-safety-checks"]
-
+default = ["obj"]
```
omni-viral: Did you mean to add it?

Veykril: whoops, didn't mean to commit that. Had that for testing with rls.

shader/src/lib.rs
```diff
@@ -27,12 +27,75 @@ pub use self::reflect::SpirvReflection;
...
+    /// An io error occured.
+    Io(std::io::Error),
+    /// Shaderc returned an error.
+    #[cfg(feature = "shader-compiler")]
+    ShaderC(::shaderc::Error),
+    /// Neither a vertex nor a compute shader has been provided.
+    NoVertComputeProvided,
+    /// A reflection error occured.
+    #[cfg(feature = "spirv-reflection")]
+    Reflect(self::reflect::ReflectError),
```
Frizi: I know this isn't your doing, but I see now how nasty this can be. Having an public enum variant under a feature flag can create all sorts of weird behaviour in the code using it. I think we can let it slip for now as this PR is presenting a pretty reasonable workaround, but we might need to take care of that eventually.

Frizi: Or actually, looking back at this, is there really a reason (apart from consistency) to have that `ReflectError` be part of `ShaderError`? The actual shader operations are pretty distinct from shader reflection in the code, I can't find a place where `ReflectError` variant of `ShaderError` would actually be returned. I might be missing it due to `Into` cast, but if not then we can probably just get rid of that variant.

Frizi:
While i'm fairly certain this makes things better, I'd still love to see the numbers :)

How does the build time compare from before and after change?

Veykril:
The time savings aren't really much apart from saving the time to build failure, since with this this PR it disappears from the build completely, which saves around 4 seconds on my machine(I've talked about the times a bit more here https://github.com/amethyst/rendy/issues/203#issuecomment-534167374).

I also just realized that I am not making the `ReflectError`, `ReflectTypeError` and `RetrievalKind` types visible, I was wondering why the doc lint didn't trigger and even more that rust doesn't disallow private types in public enums. So I'll reexport those from their module if that's okay, though it does seem to clutter the docs enum part a bit maybe.
image

shader/src/lib.rs
Veykril: Actually on a second note, I think we can remove the reflect variant and move the `NoVertComputeProvided` into the `ReflectError` since the only place where ReflectErrors in the top level module are returned is in the `reflect` function where no ShaderError is returned.
https://github.com/amethyst/rendy/pull/211/files/128305631108280991f770b4c9534bbc2af3eb57#diff-babaf69d02068afaeb320a77432c8969L368-R462

So this only leaves the Shaderc variant dependent on a feature.

Frizi: I wanted to suggest that, but realised that shaderc stays there. In that case we can't go all the way to remove the hidden option hack. I don't think there is a point.

Veykril: Ye, I saw the suggestion cause of the mail I received which is why I thought about it too then 😅. And while this doesn't solve the problem it still makes things a bit cleaner I think which is why I went through with.

omni-viral:
@Veykril The thing is. You can't have private type in public function signature, but you **can** not reexport it from private module and thus make symbol unnameable in other crates.

shader/src/lib.rs
```diff
@@ -22,17 +22,63 @@ mod reflect;
...
+pub enum ShaderError {
+    /// Shaderc could not be initialized.
+    Init,
+    /// The given path is not a valid UTF-8 string.
+    NonUtf8Path(std::path::PathBuf),
+    /// An io error occured.
+    Io(std::io::Error),
+    /// Shaderc returned an error.
+    #[cfg(feature = "shader-compiler")]
+    ShaderC(::shaderc::Error),
```
omni-viral: This have just occured to me, but this is implementation specific error, which means that `Shader` trait should have associated `Error` type it returns.

Veykril: This seems indeed better, if I understood your intention correctly. As it seems though currently none of the operations in the `lib` module seem to be able to error aside from `gfx-hal` errors so I just used an empty Error enum now, in case that behaviour is planned to change?

shader/src/shaderc.rs
```diff
@@ -36,7 +75,7 @@ where
     E: AsRef<str>,
 {
     /// Precompile shader source code into Spir-V bytecode.
-    pub fn precompile(&self) -> Result<SpirvShader, failure::Error>
+    pub fn precompile(&self) -> Result<SpirvShader, <Self as Shader>::Error>
```
omni-viral: Could you, please, specify concrete type here.

shader/src/shaderc.rs
```diff
@@ -53,21 +92,23 @@ where
     P: AsRef<std::path::Path> + std::fmt::Debug,
     E: AsRef<str>,
 {
-    fn spirv(&self) -> Result<std::borrow::Cow<'static, [u32]>, failure::Error> {
+    type Error = ShaderCError;
+
+    fn spirv(&self) -> Result<std::borrow::Cow<'static, [u32]>, <Self as Shader>::Error> {
```
omni-viral: And here ;)

shader/src/shaderc.rs
```diff
@@ -117,7 +158,7 @@ where
     E: AsRef<str>,
 {
     /// Precompile shader source code into Spir-V bytecode.
-    pub fn precompile(&self) -> Result<SpirvShader, failure::Error>
+    pub fn precompile(&self) -> Result<SpirvShader, <Self as Shader>::Error>
```
omni-viral: And here ;)

shader/src/shaderc.rs
Veykril: I actually cannot change this to the concrete type due to the `where Self: Shader` bound on precompile. Rust complains that ```
std::convert::From<<shaderc::SourceCodeShaderInfo<P, E, S> as Shader>::Error>` is not implemented for `shaderc::ShaderCError```. Should i rewrite the impl block bounds of the precompile function to be the same as the `Shader` impl block or keep it as is?

PR #213, opened 2019-09-24
Allow creating spirv shader from byte array

Frizi:


Frizi:
bors r=jaynus

Frizi:
bors retry

parasyte:
FWIW this PR introduces undefined behavior when `&[u8]` is not aligned to a 4-byte boundary. This topic came up recently in another project: https://github.com/MaikKlein/ash/pull/245#discussion_r339310161 and has been discussed at length elsewhere.

Ralith:
That's not actually the case. This calls [the `read_spirv` helper](https://github.com/gfx-rs/gfx/blob/master/src/hal/src/pso/mod.rs#L235-L290) to copy the input into a `Vec<u32>` (and fix up endianness if necessary).

parasyte:
Ohh, yes, sorry. 😐 Nothing to see here...

PR #228, opened 2019-10-22
Evolve Util into Core

omni-viral:
This crate renames `rendy-util` to `rendy-core` to better reflect what it is actually supposed to contain.
`gfx-hal` crate gets reexported through `rendy-core`.


omni-viral:
This is rather big change, but mostly just changing import path.

omni-viral:
Renaming drops coverage by 1%? Ok

PR #229, opened 2019-10-25
Add function to turn &[u8] into Vec<u32>

icefoxen:
Basically for loading SPIR-V from files, so I don't have to
bother with shaderc all the time.  Suggestions welcome!


icefoxen:
Oops, nevermind, see https://docs.rs/gfx-hal/0.3.1/gfx_hal/pso/fn.read_spirv.html

I KNEW I'd seen something like that before but couldn't find it.

PR #230, opened 2019-10-25
move examples to their own crate

### LINKED ISSUE #226: Convert examples into crates with binaries
To remove common confusion examples should be moved into `rendy-examples` crate with binary per example.

boringcactus:
fixes #226

i'm not quite sure how features should be handled in both `examples` and `rendy`; if there's a better way to handle that than what i've done, let me know

the examples don't actually run for me on dx12, both on master and after these changes, because the assert at https://github.com/gfx-rs/gfx/blob/3d5db15661127c8cad8d85522a68ec36c82f6e69/src/backend/dx12/src/pool.rs#L102 fails.

EDIT: they run for a short time on vulkan and then crash with a `STATUS_ACCESS_VIOLATION`, but that shouldn't be a result of any of this.


ezpuzz:
I find this more confusing as you can no longer use `cargo run --example triangle` which is ecosystem standard.  I believe the original confusion was just where to find the examples?  I think we could take the approach that amethyst does and instead of a top level virtual workspace use an actual top level package (rendy) and keep examples at the top of repo.

PR #231, opened 2019-10-25
Minor tweaks to shader info docs

swooster:
The biggest difference is that this fixes erroneous docs that suggested
SourceCodeShaderInfo loads source code from files, and explains why a
path must be provided.


swooster:
When reading through the updated triangle demo, I was confused about why it was providing both sourcecode and a path to the shader. Hopefully tweaks along these lines might save the next person some time.

PR #239, opened 2019-10-30
[WIP] Init

### LINKED ISSUE #224: Backend agnostic initialization
Rendy should be able to pick one of available backends and self-initialize based on preferences or config rather than caller code.

- [ ] `rendy::init` function that returns enum with one of `Rendy<B>` (usable with backend matching macro from `rendy-core`).
- [ ] `rendy::init_and_then!(pattern => expression)` where pattern accepts `Rendy<B>` value and arm expression valid for any enabled backend type.

omni-viral:
Fixes #224

### DIFF
--- a/factory/src/factory.rs
+++ b/factory/src/factory.rs
@@ -5,7 +5,7 @@ use {
             families_from_device, CommandPool, Families, Family, FamilyId, Fence, QueueType, Reset,
         },
         config::{Config, DevicesConfigure, HeapsConfigure, QueuesConfigure},
-        core::{rendy_backend_match, rendy_with_slow_safety_checks, Device, DeviceId, Instance},
+        core::{rendy_with_slow_safety_checks, Device, DeviceId, Instance, InstanceId},
         descriptor::DescriptorAllocator,
         memory::{self, Heaps, MemoryUsage, TotalMemoryUtilization, Write},
         resource::*,
@@ -17,7 +17,8 @@ use {
             adapter::{Adapter, Gpu, PhysicalDevice},
             buffer,
             device::{
-                AllocationError, Device as _, MapError, OomOrDeviceLost, OutOfMemory, WaitFor,
+                AllocationError, CreationError, Device as _, MapError, OomOrDeviceLost,
+                OutOfMemory, WaitFor,
             },
             format, image,
             pso::DescriptorSetLayoutBinding,
@@ -95,6 +96,30 @@ pub enum UploadError {
     Upload(OutOfMemory),
 }
 
+enum InstanceOrId<B: Backend> {
+    Instance(Instance<B>),
+    Id(InstanceId),
+}
+
+impl<B> InstanceOrId<B>
+where
+    B: Backend,
+{
+    fn id(&self) -> InstanceId {
+        match self {
+            InstanceOrId::Instance(instance) => instance.id(),
+            InstanceOrId::Id(id) => *id,
+        }
+    }
+
+    fn as_instance(&self) -> Option<&Instance<B>> {
+        match self {
+            InstanceOrId::Instance(instance) => Some(instance),
+            InstanceOrId::Id(_) => None,
+        }
+    }
+}
+
 /// Higher level device interface.
 /// Manges memory, resources and queue families.
 #[derive(derivative::Derivative)]
@@ -112,7 +137,7 @@ pub struct Factory<B: Backend> {
     #[derivative(Debug = "ignore")]
     adapter: Adapter<B>,
     #[derivative(Debug = "ignore")]
-    instance: Instance<B>,
+    instance: InstanceOrId<B>,
 }
 
 #[allow(unused)]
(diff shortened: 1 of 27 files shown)


init/src/windowed.rs
```diff
@@ -0,0 +1,264 @@
...
+    pub factory: Factory<B>,
+    pub families: Families<B>,
+    pub surface: Surface<B>,
+    pub window: Window,
+}
+
+impl<B: Backend> WindowedRendy<B> {
+    fn init_non_gl<T>(
+        config: &Config<impl DevicesConfigure, impl HeapsConfigure, impl QueuesConfigure>,
+        window_builder: Cow<WindowBuilder>,
```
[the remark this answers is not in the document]
omni-viral: You get a `Factory` which still has `fn create_surface`

init/src/windowed.rs
```diff
@@ -0,0 +1,264 @@
...
+            pub fn init_gl<T: 'static>(config: &Config<impl DevicesConfigure, impl HeapsConfigure, impl QueuesConfigure>, window_builder: WindowBuilder, event_loop: &EventLoop<T>) -> Result<Self, WindowedRendyInitError> {
+                use rendy_core::hal::format::AsFormat;
+
+                let windowed_context = unsafe {
+                    let builder = rendy_core::gl::config_context(
+                        rendy_core::gl::glutin::ContextBuilder::new(),
+                        rendy_core::hal::format::Rgba8Srgb::SELF,
+                        None,
+                    )
+                    .with_vsync(true);
```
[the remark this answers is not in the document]
omni-viral: Trying to keep API surface minimal here.
What would you do?

graph/src/node/render/pass.rs
```diff
@@ -718,7 +720,7 @@ where
             None
         };
 
-        let release = if !is_metal::<B>() {
+        let release = if !is_metal::<B>() && !is_gl::<B>() {
```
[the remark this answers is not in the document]
omni-viral: Makes sense.

factory/src/factory.rs
```diff
@@ -95,6 +96,30 @@ pub enum UploadError {
     Upload(OutOfMemory),
 }
 
+enum InstanceOrId<B: Backend> {
```
kvark: oh wow, this seems quite tricky, if not dirty
perhaps there needs to be a comment saying that one of these is only for GL compatibility and will hopefully be removed soon

factory/src/factory.rs
```diff
@@ -95,6 +96,30 @@ pub enum UploadError {
...
+    B: Backend,
+{
+    fn id(&self) -> InstanceId {
+        match self {
+            InstanceOrId::Instance(instance) => instance.id(),
+            InstanceOrId::Id(id) => *id,
+        }
+    }
+
+    fn instance(&self) -> Option<&Instance<B>> {
```
kvark: nit: `as_instance` ?

graph/src/node/render/pass.rs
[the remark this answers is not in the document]
kvark: I recommend switching to run-time here instead of compile-time: we may start using barriers in Metal, for example, soon

init/src/windowed.rs
```diff
@@ -0,0 +1,264 @@
...
+        }
+    }
+}
+
+impl std::error::Error for WindowedRendyInitError {}
+
+#[derive(Debug)]
+/// Initialized rendy instance with window.
+/// Create with `WindowedRendy::init`.
+pub struct WindowedRendy<B: Backend> {
```
kvark: does that mean the other `Rendy` is only for off-screen tasks?

rendy/src/lib.rs
```diff
@@ -94,3 +70,38 @@ pub use rendy_texture as texture;
 #[cfg(feature = "wsi")]
 #[doc(inline)]
 pub use rendy_wsi as wsi;
+
+/// Init rendy and execute code based on chosen backend
+#[cfg(feature = "init")]
+#[macro_export]
+macro_rules! init_and_then {
+    (($config:expr) ($factory:pat, $families:pat) => $code:block) => {{
+        match $crate::init::AnyRendy::init_auto($config) {
```
kvark: this part isn't backend-dependent (`AnyRendy` creation), does it have to be in a macro?

init/src/windowed.rs
omni-viral: No. Other one created without surface, but if backend supports it - surface can be created.

graph/src/node/render/pass.rs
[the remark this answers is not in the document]
omni-viral: How would I know in runtime if metal backend started using barriers?

init/src/windowed.rs
omni-viral: No. Other one created without surface, but if backend supports it - surface can be created.

rendy/src/lib.rs
omni-viral: Otherwise usage would look like
```
rendy::with_any_rendy!(AnyRendy::init_auto(config)) (factory, families) => { ... });
```
Well. Maybe that's better

graph/src/node/render/pass.rs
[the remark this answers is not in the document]
kvark: Not exposed today, but we can have a "hint" set of flags, similar in spirit to https://github.com/gfx-rs/gfx/pull/3042
cc @aleksijuvani

PR #245, opened 2019-11-14
Add Error impl for BuildError

valkum:
BuildError is missing an Error implementation.

I was not sure about the cause for Format as Format is no Error. Advice would be nice here.
The rest follows the error impls from the latest commits.


termhn:
Need this for amethyst and rendy-pbr, so.

bors r+

PR #250, opened 2019-11-28
[0.5] Fix memory invalidation logic

kvark:
Sibling of #249 for the published version
cc @zakorgy


kvark:
@omni-viral publish please!

PR #251, opened 2019-12-01
Dynamic meshes

omni-viral:


core/src/casts.rs
```diff
@@ -1,24 +1,46 @@
...
 pub fn cast_slice<T>(slice: &[T]) -> &[u8] {
-    let len = std::mem::size_of::<T>() * slice.len();
+    let len = size_of::<T>() * slice.len();
     let ptr = slice.as_ptr();
     unsafe { std::slice::from_raw_parts(ptr as _, len) }
 }
 
+/// Cast slice of some arbitrary type into slice of bytes.
+pub unsafe fn cast_arbitrary_slice<T, U>(slice: &[T]) -> &[U] {
+    let bytes = size_of::<T>() * slice.len();
```
Frizi: This is terribly unsafe in so many ways, a comment should definitely mention that it's effectively equivalen to transmute. You are only using it on `Copy` types, so it should probably be typeguarded as such. That way we at least get rid of `Drop` related issues.

mesh/src/dynamic.rs
```diff
@@ -0,0 +1,624 @@
...
+                let ptr = ptr as *mut V;
+                let additional = additional_bytes / size_of::<V>();
+                for ptr in std::iter::successors(Some(ptr), |p| Some(p.add(1))).take(additional) {
+                    std::ptr::write(ptr, fill);
+                }
+                vertices.bytes.set_len(bytes_range.end);
+            }
+        }
+
+        Ok(unsafe { cast_arbitrary_slice_mut(&mut vertices.bytes[bytes_range]) })
```
Frizi: I think this is a safety issue. We probably should annotate `AsVertex` trait as unsafe so it is a responsibility of implementer to make sure it's safe to represent it as a byte slice (basically enforce `#[repr(C)]`).

mesh/src/dynamic.rs
```diff
@@ -0,0 +1,624 @@
...
+    /// If requested range is larger than vertex array then it will be resized, filling new values with `fill` before returning.
+    ///
+    /// # Panics
+    ///
+    /// This function will panic if wrong vertex type is requested.
+    pub fn modify_vertices<V>(
+        &mut self,
+        index: usize,
+        range: Range<u32>,
+        fill: V,
```
Frizi: I would rename `fill` to `default_value` or something like that. I interpreted those arguments as if the function would fill the whole provided range with `fill` value.

mesh/src/dynamic.rs
```diff
@@ -0,0 +1,624 @@
...
+    }
+
+    /// Acquire slice of indices from index buffer at `index`.
+    /// All indices from specifed range will be marked dirty and flushed on `DynamicMesh::update`.
+    /// If requested range is larger than index array then it will be resized, filling new values with `0` before returning.
+    ///
+    /// # Panics
+    ///
+    /// This function will panic if wrong index type is requested.
+    pub fn modify_indices_16(
```
Frizi: It seems to me that it would be better to have a single `modify_indices` that returns either `&mut [u16]` or `&mut [u32]`, then possibly have those two methods use it. There is quite some duplication here.

Also, doc is inaccurate. It doesn't panic, it returns an error on mismatch.

mesh/src/dynamic.rs
```diff
@@ -0,0 +1,624 @@
...
+                let ptr = indices.bytes.as_mut_ptr().add(indices.bytes.len());
+                std::ptr::write_bytes(ptr, 0, additional_bytes);
+                indices.bytes.set_len(bytes_range.end);
+            }
+        }
+
+        Ok(unsafe { cast_arbitrary_slice_mut(&mut indices.bytes[bytes_range]) })
+    }
+
+    pub fn update(&mut self, queue: QueueId, factory: &Factory<B>) -> Result<(), UploadError> {
```
Frizi: i think this should be called `upload` or `flush`. Also the upload should only really happen if it was updated since last time (it should have a dirty flag or something).

core/src/casts.rs
omni-viral: Also intended as internal method. Shouldn't be pub.

mesh/src/dynamic.rs
omni-viral: It is always safe to represent any type as bytes.

omni-viral: And should be also safe to represent back. Alignment should be preserved.

mesh/src/dynamic.rs
omni-viral: It **does** use dirty regions

Frizi: Oh shit, you are totally right. I somehow skipped that in reading. Then it's only the naming.

mesh/src/dynamic.rs
Frizi: No, it is not always safe. See this thread https://internals.rust-lang.org/t/pre-rfc-v2-safe-transmute/11431

In fact, there are crates that help with this today. See [bytemuck](https://docs.rs/bytemuck/1.1.0/bytemuck/). Notice that `Pod` trait is actually unsafe and has quite some safety rules attached to it.

PR #252, opened 2019-12-02
Add Encoder::copy_image_to_buffer

fintelia:
The crate currently provides `copy_buffer_to_image` as well as `copy_buffer` and `copy_image`, but the last version is missing.


fintelia:
The CI failure seems to be unrelated to this patch (something about not having enough disk space to clone the repository on MacOS?)

Issue #201, opened 2019-09-05
Respect nonCoherentAtomSize and align blocks mapping internally.


omni-viral:

kvark:
Getting the following validation error in an wgpu-based application with rendy-memory-0.5.2:
> 447 API High Miscellaneous 687 vkFlushMappedMemoryRanges: Offset in pMemRanges[0] is 0x203, which is not a multiple of VkPhysicalDeviceLimits::nonCoherentAtomSize (0x40). The Vulkan spec states: offset must be a multiple of VkPhysicalDeviceLimits::nonCoherentAtomSize (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-VkMappedMemoryRange-offset-00687)

PR #257, opened 2020-01-14
Fix problem with non-coherent memory mapping

omni-viral:
Fixes: https://github.com/amethyst/rendy/issues/201

### DIFF
--- a/memory/src/allocator/dynamic.rs
+++ b/memory/src/allocator/dynamic.rs
@@ -206,18 +206,18 @@ where
             config
         );
 
+        assert!(
+            config.block_size_granularity.is_power_of_two(),
+            "Allocation granularity must be power of two"
+        );
+
         let block_size_granularity = if is_non_coherent_visible(memory_properties) {
             non_coherent_atom_size
-                .next_power_of_two()
                 .max(config.block_size_granularity)
+                .next_power_of_two()
         } else {
             config.block_size_granularity
         };
-
-        assert!(
-            block_size_granularity.is_power_of_two(),
-            "Allocation granularity must be power of two"
-        );
 
         assert!(
             config.max_chunk_size.is_power_of_two(),
@@ -555,6 +555,14 @@ where
         debug_assert!(size <= self.max_allocation());
         debug_assert!(align.is_power_of_two());
         let aligned_size = ((size - 1) | (align - 1) | (self.block_size_granularity - 1)) + 1;
+
+        // This will change nothing if `self.non_coherent_atom_size` is power of two.
+        // But just in case...
+        let aligned_size = if is_non_coherent_visible(self.memory_properties) {
+            align_size(aligned_size, self.non_coherent_atom_size)
+        } else {
+            aligned_size
+        };
 
         log::trace!(
             "Allocate dynamic block: size: {}, align: {}, aligned size: {}, type: {}",
--- a/memory/src/allocator/dynamic.rs
+++ b/memory/src/allocator/dynamic.rs
@@ -79,18 +79,30 @@ where
             range.start < range.end,
             "Memory mapping region must have valid size"
         );
+
         if !self.shared_memory().host_visible() {
             //TODO: invalid access error
             return Err(gfx_hal::device::MapError::MappingFailed);
         }
 
+        let requested_range = relative_to_sub_range(self.range.clone(), range)
+            .ok_or(gfx_hal::device::MapError::OutOfBounds)?;
+
+        let mapping_range = if !self.shared_memory().host_coherent() {
+            align_range(
+                requested_range.clone(),
+                self.shared_memory().non_coherent_atom_size(),
+            )
+        } else {
+            requested_range.clone()
+        };
+
         if let Some(ptr) = self.ptr {
-            if let Some((ptr, range)) = mapped_sub_range(ptr, self.range.clone(), range) {
-                let mapping = unsafe { MappedRange::from_raw(self.shared_memory(), ptr, range) };
-                Ok(mapping)
-            } else {
-                Err(gfx_hal::device::MapError::OutOfBounds)
-            }
+            let ptr = mapped_sub_range(ptr, self.range.clone(), mapping_range.clone()).unwrap();
+            let mapping = unsafe {
+                MappedRange::from_raw(self.shared_memory(), ptr, mapping_range, requested_range)
+            };
+            Ok(mapping)
         } else {
             Err(gfx_hal::device::MapError::MappingFailed)
         }
@@ -140,6 +152,7 @@ pub struct DynamicAllocator<B: Backend> {
 
     /// Ordered set of sizes that have allocated chunks.
     chunks: BTreeSet<u64>,
+    non_coherent_atom_size: u64,
 }
 
 unsafe impl<B> Send for DynamicAllocator<B> where B: Backend {}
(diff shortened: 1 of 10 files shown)


memory/src/allocator/dynamic.rs
```diff
@@ -79,18 +79,30 @@ where
...
             return Err(gfx_hal::device::MapError::MappingFailed);
         }
 
+        let requested_range = relative_to_sub_range(self.range.clone(), range)
+            .ok_or(gfx_hal::device::MapError::OutOfBounds)?;
+
+        let mapping_range = if !self.shared_memory().host_coherent() {
+            align_range(
+                requested_range.clone(),
+                self.shared_memory().non_coherent_atom_size(),
```
kvark: nit: perhaps it would look nice if we make this method return an option, we'd then have
```rust
let mapping_range = match self.shared_memory().atom_size() {
  Some(atom_size) => align_range(requested_range.clone(), atom_size),
  None => requested_range.clone(),
};
```

memory/src/allocator/dynamic.rs
```diff
@@ -79,18 +79,30 @@ where
...
+        };
+
         if let Some(ptr) = self.ptr {
-            if let Some((ptr, range)) = mapped_sub_range(ptr, self.range.clone(), range) {
-                let mapping = unsafe { MappedRange::from_raw(self.shared_memory(), ptr, range) };
-                Ok(mapping)
-            } else {
-                Err(gfx_hal::device::MapError::OutOfBounds)
-            }
+            let ptr = mapped_sub_range(ptr, self.range.clone(), mapping_range.clone()).unwrap();
```
kvark: Would it be simpler to accept all the ranges by reference in the helper methods? They don't really need the values, and having `clone()` everywhere is noisy

memory/src/mapping/mod.rs
```diff
@@ -35,7 +33,11 @@ pub struct MappedRange<'a, B: Backend, C = MaybeCoherent> {
     ptr: NonNull<u8>,
 
     /// Range of mapped memory.
-    range: Range<u64>,
+    mapping_range: Range<u64>,
```
kvark: Do we have to store both? It seems to me that we could reconstruct the `mapping_range` in most (if not all) cases easily

memory/src/mapping/mod.rs
```diff
@@ -86,14 +88,37 @@ where
...
     /// `range` is in memory object space.
     /// `ptr` points to the `range.start` offset from memory origin.
-    pub unsafe fn from_raw(memory: &'a Memory<B>, ptr: NonNull<u8>, range: Range<u64>) -> Self {
-        assert!(
-            range.start < range.end,
+    pub(crate) unsafe fn from_raw(
+        memory: &'a Memory<B>,
+        ptr: NonNull<u8>,
+        mapping_range: Range<u64>,
+        requested_range: Range<u64>,
```
kvark: now we technically have 3 kinds of ranges:
  - given range (coming from the user), which is relative to their view of memory
  - absolute range based off the *real* allocation, corresponds to `requested range`` here
  - aligned absolute range, corresponds to `mapping_range`

Ideally, we'd want to separate them at the type level.

memory/src/util.rs
```diff
@@ -123,3 +123,35 @@ pub(crate) fn fits_usize<T: IntegerFitting>(value: T) -> bool {
 pub(crate) fn fits_u32(value: usize) -> bool {
     u32::usize_fits(value)
 }
+
+pub(crate) fn align_range(range: std::ops::Range<u64>, align: u64) -> std::ops::Range<u64> {
+    let start = range.start - range.start % align;
```
kvark: is `align` guaranteed to be a power of 2? we could make this code faster for this case

memory/src/util.rs
```diff
@@ -123,3 +123,35 @@ pub(crate) fn fits_usize<T: IntegerFitting>(value: T) -> bool {
...
+    let end = ((range.end - 1) / align + 1) * align;
```
kvark: where are we handling the case of `range.end == 0`?

memory/src/util.rs
```diff
@@ -123,3 +123,35 @@ pub(crate) fn fits_usize<T: IntegerFitting>(value: T) -> bool {
...
+    start..end
+}
+
+pub(crate) fn align_size(size: u64, align: u64) -> u64 {
+    ((size - 1) / align + 1) * align
```
kvark: same here

memory/src/util.rs
```diff
@@ -123,3 +123,35 @@ pub(crate) fn fits_usize<T: IntegerFitting>(value: T) -> bool {
...
+}
+
+pub(crate) fn is_non_coherent_visible(properties: gfx_hal::memory::Properties) -> bool {
+    properties.contains(gfx_hal::memory::Properties::CPU_VISIBLE)
```
kvark: this is unlikely a performance-sensitive piece, but if we wanted it to be super nice, we'd want to do something like
```rust
properties & (gfx_hal::memory::Properties::CPU_VISIBLE | gfx_hal::memory::Properties::COHERENT) == gfx_hal::memory::Properties::CPU_VISIBLE
```

memory/src/util.rs
```diff
@@ -123,3 +123,35 @@ pub(crate) fn fits_usize<T: IntegerFitting>(value: T) -> bool {
...
+}
+
+pub(crate) fn relative_to_sub_range(
+    range: std::ops::Range<u64>,
+    relative: std::ops::Range<u64>,
+) -> Option<std::ops::Range<u64>> {
+    let start = relative.start + range.start;
+    let end = relative.end + range.start;
+    if end <= range.end {
+        Some(start..end)
```
kvark: in what legitimate case this wouldn't be true?

memory/src/util.rs
```diff
@@ -123,3 +123,35 @@ pub(crate) fn fits_usize<T: IntegerFitting>(value: T) -> bool {
...
+    ((size - 1) / align + 1) * align
+}
+
+pub(crate) fn is_non_coherent_visible(properties: gfx_hal::memory::Properties) -> bool {
+    properties.contains(gfx_hal::memory::Properties::CPU_VISIBLE)
+        && !properties.contains(gfx_hal::memory::Properties::COHERENT)
+}
+
+pub(crate) fn relative_to_sub_range(
+    range: std::ops::Range<u64>,
```
kvark: this is a tad confusing: the function starts with `relative_to`, so I'd expect the relative range to be the first argument

memory/src/allocator/dynamic.rs
omni-viral: It would be simpler to use custom `Range` that is `offset + size` and `Copy`.
But this is out of scope for this bugfix. Let's put it into tech debt.

memory/src/mapping/mod.rs
omni-viral: I drafted that. But my rendy time ran out so I pushed without those.
I will return to it later.

memory/src/util.rs
omni-viral: Spec says `nonCoherentAtomSize is the size and alignment in bytes that bounds concurrent access to host-mapped device memory.` so I have to assume it may be not power of 2.

Which means there's a bug where I assume that next power of two is divisible by `nonCoherentAtomSize`...

memory/src/util.rs
omni-viral: All public functions that accept ranges assert that `range.start < range.end`

omni-viral: I could insert `debug_assert_ne!(range.end, 0)`

azriel91:
@omni-viral Heya, would you be able to find time to continue on this? (seeing that it looks nearly done)

kvark:
@omni-viral ping!

kvark:
@omni-viral would be really nice to have this finished!

Issue #266, opened 2020-02-06
Generate vertex tangents for OBJ files


GrantMoyer:
Currently, `rendy::mesh::obj::load_from_obj()` does not generate vertex tangents when loading OBJ files. This prevents OBJ files from being used with Amethyst's RenderBundle's RenderPbr3D plugin, which needs vertex tangents.

I can implement this, but I want to make sure this is a desired feature, and if so, what tradeoffs are preferred before I start. There are 3 ways I see to implement this:

1. Just generate the tangents all the time
  * Wastes CPU cycles if tangents aren't needed
2. Add a boolean parameter to specify if tangents are desired
  * Breaks API
  * Inflexible; tangents treated as more special than other vertex attributes
3. Make `load_from_obj` generic over `Vec<A> where A: AsVertex` and tuples of those types and add implementations for the relevant types (Vec<Position>, Vec<Normal>, etc.)
  * Breaks API
  * Leaves room for future extension
  * Adds types complexity

My intuition is that 1 is the best option if OBJ loading time is considered negligible, and 3 is the best option otherwise.

GrantMoyer:
Related issue: https://github.com/amethyst/amethyst/issues/2104

PR #267, opened 2020-02-08
Implement tangent computation for OBJ loading

GrantMoyer:
Issue #266 

I went ahead and implemented indiscriminate tangent computation during OBJ file loading. I should be relatively straightforward to convert it to one of the other options mention in the issue if that's desirable.

### DIFF
--- a/mesh/src/format/obj.rs
+++ b/mesh/src/format/obj.rs
@@ -2,7 +2,8 @@
 
 use log::trace;
 use {
-    crate::{mesh::MeshBuilder, Normal, Position, TexCoord},
+    crate::{mesh::MeshBuilder, Normal, Position, Tangent, TexCoord},
+    smallvec::SmallVec,
     wavefront_obj::obj,
 };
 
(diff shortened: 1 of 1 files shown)


GrantMoyer:
This effectively loads per face tangents, since it computes tangents per tri without respect to neighboring tri's. The most straightforward way to compute per vertex tangents probably involves implementing proper indexing instead of duplicating shared vertices.

GrantMoyer:
Okay, this should now compute per vertex tangents. They should still probably be weighted by face area or some other method.

termhn:
You could use this crate, mikktspace is pretty standard as a very effective
tangent space now. https://github.com/gltf-rs/mikktspace

GrantMoyer:
Thanks for the pointer. I was surprised how non-standardized tangent space computation was when I looked into it. I'm marking this ready for review now.

GrantMoyer:
Is there anything blocking this that I could fix or help with?

PR #265, opened 2020-01-26
Update examples with winit 0.20.0

LiardeauxQ:
Examples not compiled with winit 0.20.0 so I replaced no longer existing functions and renamed elements.

PR #268, opened 2020-02-15
Updated dependency versions.

azriel91:
Bumped dependencies, and updated `rendy/examples/meshes` to test that it still works.


azriel91:
Incorporated changes from #265. However I'm sure I'm doing something bad<sup>TM</sup> in `image.rs` -- I don't understand the image formats, so simply made it compile.

The sprite example runs though:

```bash
cargo run --example sprite --features="base init-winit texture-image shader-compiler vulkan"
```

PR #269, opened 2020-02-17
Update Rendy Descriptor For Latest Hal?

zicklag:
I have absolutely no idea what I'm doing, but I was just messing around trying to get `wgpu-rs` to compile with OpenGL support and this got in the way because it seems it is not up-to-date with the latest `gfx-hal`.

Is this correct or is there a better way to do this? If I'm way off this may not be worth your time and you can feel free to close this. :slightly_smiling_face: 

Related: https://github.com/gfx-rs/wgpu/issues/450

**Edit:** I'll squash the commits before merging if this looks good.

### DIFF
--- a/descriptor/src/ranges.rs
+++ b/descriptor/src/ranges.rs
@@ -3,23 +3,114 @@ use std::{
     ops::{Add, AddAssign, Mul, MulAssign, Sub, SubAssign},
 };
 
-pub use gfx_hal::pso::{DescriptorRangeDesc, DescriptorSetLayoutBinding, DescriptorType};
+pub use gfx_hal::pso::{
+    BufferDescriptorFormat, BufferDescriptorType, DescriptorRangeDesc, DescriptorSetLayoutBinding,
+    DescriptorType, ImageDescriptorType,
+};
 
 const DESCRIPTOR_TYPES_COUNT: usize = 11;
 
 const DESCRIPTOR_TYPES: [DescriptorType; DESCRIPTOR_TYPES_COUNT] = [
     DescriptorType::Sampler,
-    DescriptorType::CombinedImageSampler,
-    DescriptorType::SampledImage,
-    DescriptorType::StorageImage,
-    DescriptorType::UniformTexelBuffer,
-    DescriptorType::StorageTexelBuffer,
-    DescriptorType::UniformBuffer,
-    DescriptorType::StorageBuffer,
-    DescriptorType::UniformBufferDynamic,
-    DescriptorType::StorageBufferDynamic,
+    DescriptorType::Image {
+        ty: ImageDescriptorType::Sampled { with_sampler: true },
+    },
+    DescriptorType::Image {
+        ty: ImageDescriptorType::Sampled {
+            with_sampler: false,
+        },
+    },
+    DescriptorType::Image {
+        ty: ImageDescriptorType::Storage,
+    },
+    DescriptorType::Buffer {
+        ty: BufferDescriptorType::Storage { read_only: false },
+        format: BufferDescriptorFormat::Structured {
+            dynamic_offset: true,
+        },
+    },
+    DescriptorType::Buffer {
+        ty: BufferDescriptorType::Uniform,
+        format: BufferDescriptorFormat::Structured {
+            dynamic_offset: true,
+        },
+    },
+    DescriptorType::Buffer {
+        ty: BufferDescriptorType::Storage { read_only: false },
+        format: BufferDescriptorFormat::Structured {
+            dynamic_offset: false,
+        },
+    },
+    DescriptorType::Buffer {
+        ty: BufferDescriptorType::Uniform,
+        format: BufferDescriptorFormat::Structured {
+            dynamic_offset: false,
+        },
+    },
+    DescriptorType::Buffer {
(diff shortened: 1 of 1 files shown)


descriptor/src/ranges.rs
```diff
@@ -3,24 +3,149 @@ use std::{
...
+    DescriptorType::Image {
+        ty: ImageDescriptorType::Sampled {
+            with_sampler: false,
+        },
+    },
+    DescriptorType::Image {
+        ty: ImageDescriptorType::Storage,
+    },
+    DescriptorType::Buffer {
+        ty: BufferDescriptorType::Storage { read_only: true },
```
kvark: from Rendy perspective, `read_only` means nothing, so we don't need to consider this flag as a difference here

zicklag: Fix it and pushed updates.

zicklag:
I left out the update to the `Cargo.toml` needed to make this work. How do you want that done? What commit should I point it at, and do the rest of the Rendy creates need to be updated at the same time?

zicklag:
Ah, I just realized that this isn't necessary for https://github.com/gfx-rs/wgpu/issues/450 because we are basing that work off of the `hal-0.4` branch of `gfx-hal` as opposed to the `master` branch.

kvark:
Yeah, we don't have to update Rendy for your case. Although, doing it (and updating Cargo.toml) is still useful to ease the future migration of Rendy to hal-0.5 when it's out.

glalonde:
Ping, @zicklag could you merge this? I'd like to be able to run wgpu-rs with an updated gfx-hal, which seems to require this. Thanks for getting it to this point : )

zicklag:
Needs a review from @kvark first, and need to squash the commits. I should be able to squash the commits today or tomorrow.

descriptor/src/ranges.rs
```diff
@@ -3,24 +3,115 @@ use std::{
...
-    DescriptorType::CombinedImageSampler,
-    DescriptorType::SampledImage,
-    DescriptorType::StorageImage,
-    DescriptorType::UniformTexelBuffer,
-    DescriptorType::StorageTexelBuffer,
-    DescriptorType::UniformBuffer,
-    DescriptorType::StorageBuffer,
-    DescriptorType::UniformBufferDynamic,
-    DescriptorType::StorageBufferDynamic,
+    DescriptorType::Image {
```
kvark: The danger here is missing any of the variants, especially if the upstream changes. E.g. a new variant is added, and this code still compiles.
I was thinking that a rough way to do so would be to have a test that generates, say 1000 of the *random* descriptor types and ensures that all of them are within the list.

kvark: if we are crazy about it, we could even make a procedural macro that enumerates all the variants and builds that kind of a static array :)

descriptor/src/ranges.rs
```diff
@@ -65,7 +156,7 @@ impl DescriptorRanges {
         let mut descs = Self::zero();
 
         for binding in bindings {
-            descs.counts[binding.ty as usize] += binding.count as u32;
```
kvark: since we are at it, would be useful to make this method to accept something iteratable over `DescriptorSetLayoutBinding` instead of a slice specifically

descriptor/src/ranges.rs
omni-viral: Could gfx-hal export `const DESCRIPTOR_TYPE_COUNT: usize`?

omni-viral: `gfx-hal` could export `DESCRIPTOR_TYPES_COUNT` for us.
But in any case `descriptor_type_index` would fail to compile if new variant is added.

kvark: I don't think gfx-hal should do this. It doesn't have any way of verifying that `DESCRIPTOR_TYPES_COUNT` is correct, and it doesn't have any use for it either.

zicklag: I thought about a macro, it could be a good idea. Like @omni-viral said, it is useful that the `match` in `descriptor_type_index` wouldn't compile if we didn't match on every possible variant. The risk is if the `DESCRIPTOR_TYPES` list got unsynchronized with `descriptor_type_index`.

I think I'll mess with a macro for automating the generation of the list and the function so that they stay in sync. It would allow allow us to expose `DESCRIPTOR_TYPES_COUNT` because it would automatically be kept up-to-date.

descriptor/src/ranges.rs
zicklag: That should be easy.

descriptor/src/ranges.rs
kvark: fwiw, I don't think this PR should be blocked on that. Could be a follow-up

zicklag: :+1: I'll work on those updates in a separate PR. I think this is good to go, then.

kvark:
@zicklag ping

zicklag:
Oh, I think this is ready. If I get to it I'll look into the iterator instead of slice, but that could be another PR and I can look at a macro for the matching at the same time.

kvark:
Great! @omni-viral can we proceed?

kvark:
Related to #275

kvark:
@omni-viral what's blocking this?

PR #273, opened 2020-03-19
Match amethyst `rand_core` dependencies

caitp:
`palette` includes an earlier version of rand/rand_core, which is incompatible with the version used in Amethyst. This breaks example code (using the following toolchains)

stable-x86_64-apple-darwin - Up to date : 1.42.0 (b8cedc004 2020-03-09)
nightly-x86_64-apple-darwin - Up to date : 1.44.0-nightly (7ceebd98c 2020-03-17)

This also needs an update to the palette deps in amethyst and amethest_rendy


caitp:
This turned out not to be needed, deleting and rebuilding Cargo.lock in the amethyst repo was enough.

PR #281, opened 2020-09-07
Remove wgpu-rs from users in README.md

rishflab:
wgpu-rs is not using rendy as stated in the readme

see:
https://github.com/gfx-rs/wgpu-rs/blob/master/Cargo.toml

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,6 @@ enabled separately (enabling a subcrate will also enable its dependencies).
 ## Who is using it?
 
 * The [`Amethyst`](https://github.com/amethyst/) project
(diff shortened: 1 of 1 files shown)


zicklag:
Yeah, now they are using a fork of `rendy_memory` and `rendy_descriptor` or something which was migrated to [`gfx_extras`](https://github.com/gfx-rs/gfx-extras) if I understood correctly.

Not sure if that is worth any notes. :man_shrugging:

ezpuzz:
I don't think the provenance of the forked code is that notable.

PR #314, opened 2021-06-19
Update wavefront_obj dependency

jojolepro:


zakarumych:
Isn't this a breaking change?

zakarumych:
I guess I still got rights for publishing on crates.io

jojolepro:
Gotcha

jojolepro:
Fixed! Also, do you think it would be possible to add the amethyst org/group as a crate owner, so that we can push small updates like this one if you are not available?

zakarumych:
Amethyst group is owner already.

jojolepro:
It gave access to the amethyst:rendy team, which has sadly been deleted during a cleanup. Recreating the team doesn't give access back to the crate :joy:

zakarumych:
To which team should I transfer access rights?

jojolepro:
@amethyst/crate_access
https://github.com/orgs/amethyst/teams/crate_access

zakarumych:
I've added that team to owners list

